PR #86, opened 2026-02-12
Feature/logger base version

AlexanderJarnstromBTH:


src/logger/api/src/log_api/api.py
```diff
@@ -0,0 +1,71 @@
...
+from typing import Any, Sequence
+from datetime import datetime
+
+import uvicorn
+from fastapi import FastAPI, Request
+from fastapi.exceptions import RequestValidationError
+from fastapi.responses import JSONResponse
+from fastapi.encoders import jsonable_encoder
+
+from .handlers import handel_get_logs, handel_add_log
```
jaen22BTH: I believe this will point directly to the file (currently in the same dir). However, considering we create a package structure, we can point to the 'hadel_get_logs; and 'gandel_add_log' in the actual package, by doing (log_api.handelers import handel_get_logs' ...), if you add package-dir in pyprject.toml (maybe this is done by default, idk).

src/logger/api/src/log_api/database.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+
+    if not isinstance(row, tuple):
+        raise TypeError("Row is of wrong type.")
+
+    log_id: int | None = row[0] if isinstance(row[0], int) else None
+    occured: datetime | None = row[1] if isinstance(row[1], datetime) else None
+    message: str | None = row[2] if isinstance(row[2], str) else None
+    event_type: str | None = row[3] if isinstance(row[3], str) else None
+    service: str | None = row[4] if isinstance(row[4], str) else None
```
jaen22BTH, line 37: Could the row[4] be a problem, meaning, do we always know for certain that this index will exist in the return from the DB query?

jaen22BTH (approved):

PR #173, opened 2026-02-27
Feature/frontend setup/refining

EmEhPa:
dependabot[bot] was complaining and suggested half a improvement, this should be the whole, hopefully it works.


AlexanderJarnstromBTH (approved):
crazy

PR #196, opened 2026-03-04
added boto3 as a dependecy

AlexanderJarnstromBTH:


src/connectors/gitlabs/setup.py
```diff
@@ -9,6 +9,7 @@
     install_requires=[
         "fastapi[standard]>=0.128",
         "requests>=2.32",
+        "boto3>=1.42",
```
jaen22BTH: This should not be needed as a dependency here. It should only be defined in shared_functions.

PR #198, opened 2026-03-05
Role based access control

kebabohvo:
We did:

Role-based access control meaning, we have now two roles of users that can access the dms frontend; admin & other (base user).
We implemented:
- visibility in sidebar, so that users only see what they are supposed to see.
- access, we made it so that users of different roles can't access outside of their intended pages.
- added error pages
- removed comments and testing sections


AlexanderJarnstromBTH (approved):
crazy?

PR #200, opened 2026-03-06
Fix/gitlab connector

AlexanderJarnstromBTH:


src/connectors/gitlabs/src/interfacer.py
```diff
@@ -166,11 +166,16 @@ def files_to_index(self, subdata: str | None = None) -> dict:
             branch = project.get("default_branch")
             url = f"{project.get('web_url')}/-/archive/{branch}/{project.get('path')}-{branch}.zip?ref_type=heads"
             content = requests.get(url, timeout=120).content
-            files_data.extend(self._unpack_zip(content))
+            file_pointers = self.get_files_in_project(project_id)
+            files = self._unpack_zip(content)
 
-        generated_subdata = base64.urlsafe_b64encode(json.dumps(current_subdata).encode()).decode()
+            # pleace dont keep this
+            for i, file in enumerate(files):
```
jaen22BTH, line 173: Do we really want to make the assumtion that files from ziped content always are the sorted the same way as files returned by the get_files_in_project method. I.e, even if that might seem true in todays case, do we artificially want to make this an absolute truth, or migrate to something deterministic?

src/shared_functions/pyproject.toml
```diff
@@ -8,6 +8,7 @@ version = "0.1.0"
 description = "Library containing shared functions used in DMIS."
 dependencies = [
     "requests>=2.32",
+    "boto3"
```
jaen22BTH, line 11: Do we want to pick a minimum version for this?

src/connectors/gitlabs/src/interfacer.py
AlexanderJarnstromBTH: temporary fix, go ahead if there is a more permanent one.

src/shared_functions/pyproject.toml
AlexanderJarnstromBTH: definitely

PR #208, opened 2026-03-10
mege frontend-mainapi functionality to frontend/main

kebabohvo:
We made some changes and now it works.


EmEhPa:
Woopwoop it passed all checks :D

PR #209, opened 2026-03-11
Feature/search engine

AlexanderJarnstromBTH:


src/search_engine/src/se_api/services/connector.py
```diff
@@ -83,72 +62,65 @@ def get_file(self, pointer: str, session: Session | None = None) -> File | None:
...
+        if file_url is None:
+            return []
+
+        response = get(file_url, timeout=120).json()
+
+        data = response.get("files")
+
+        if data is None:
+            dms_error("No files in collector response.")
+            return []
```
FilipDar, line 94: maybe not needed?

AlexanderJarnstromBTH: du ba llalar

jaen22BTH (approved):

PR #217, opened 2026-03-12
Feature/frontend/session sync

muaathaw:
Updated the logout flow to ensure session termination is synchronized across all open browser tabs. This prevents "ghost sessions" where a user remains partially authenticated in secondary tabs after logging out.

Changes:

Added localStorage event trigger in handleLogout.

Implemented a global storage listener in App.vue to clear sessionStorage and redirect on logout events.

Updated ErrorStatusView to handle 401 redirects properly.

### DIFF
--- a/src/api/src/dmis_api/api.py
+++ b/src/api/src/dmis_api/api.py
@@ -3,19 +3,20 @@
 from typing import Any
 import argparse
 import requests
-#from pathlib import Path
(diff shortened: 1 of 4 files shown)


EmEhPa (approved):
We will look att som error statuses later, eg. currently all 401:s is redirected to login.

PR #224, opened 2026-03-20
Feature/comment frontend code

EmEhPa:
Mostly just comments :) 
Hopefully they are good enough. Maybe not the best but it easier to understand the code now I hope.

### DIFF
--- a/src/front_end/nginx.conf.template
+++ b/src/front_end/nginx.conf.template
@@ -24,7 +24,7 @@ server {
 
     # Proxy API calls to backend (avoids CORS — same-origin from browser's view)
     location /api/ {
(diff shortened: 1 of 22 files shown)


src/front_end/nginx.conf.template
```diff
@@ -24,7 +24,7 @@ server {
 
     # Proxy API calls to backend (avoids CORS — same-origin from browser's view)
     location /api/ {
-        proxy_pass https://${API_HOST}/;
+        proxy_pass http://${API_HOST}/;
```
jaen22BTH: @jesperlenn would we rather want to specify this in the environment variable?

jaen22BTH (approved):

PR #230, opened 2026-03-23
Remove hardcode

EmEhPa:


jesperlenn:
Update names in .env, don't use VITE

src/front_end/src/views/SearchView.vue
jesperlenn, line 1: VITE_API_BASE_URL still present

jesperlenn (approved):
It works on my machine

PR #243, opened 2026-03-24
removed package-lock, it should not be a part of the files at github.…

EmEhPa:
… changed some exports to not export and removed two unsed dependencies


jesperlenn (approved):
LGTM

PR #255, opened 2026-03-25
Soure label

EmEhPa:
Now its possible to show where the file is from, currently only gitlab, but it is there

label


AlexanderJarnstromBTH (approved):
crazy

PR #258, opened 2026-03-26
Fix: Search filter now correctly filters results by type, source and security

muaathaw:
Refactored SearchFiltersCard and SearchView to correctly apply filters on search results.

Changes:
- Added allMatches ref to preserve original results
- Filters now match against filename extension, source and security fields
- Clear all button resets filters and restores full result set
- Static filter options for source, type and security classification

### DIFF
--- a/src/front_end/src/components/SearchFiltersCard.vue
+++ b/src/front_end/src/components/SearchFiltersCard.vue
@@ -1,45 +1,54 @@
 <script setup>
-import { computed, ref } from 'vue'
+import { computed, watch, reactive } from 'vue'
 import { Grid2X2, FileText, Shield } from 'lucide-vue-next'
 
-const sourceFilters = ['GitHub', 'GitLab', 'Network File System'] // Add more sources needed if possible
+const sourceFilters = ['GitHub', 'GitLab', 'Network File System']
 const typeFilters = ['PDF (.pdf)', 'Word (.docx)', 'Excel (.xlsx)', 'Text / Markdown (.txt, .md)']
 const securityFilters = ['Public', 'Internal', 'Sensitive', 'Confidential']
 
-const selectedFilters = ref({
-  source: [],
-  type: [],
-  security: []
+const props = defineProps({
+  selectedFilters: Object
 })
+const emit = defineEmits(['update:filters'])
+
+// Make reactive local copy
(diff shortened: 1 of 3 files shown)


src/front_end/src/components/SearchFiltersCard.vue
```diff
@@ -1,45 +1,54 @@
 <script setup>
-import { computed, ref } from 'vue'
+import { computed, watch, reactive } from 'vue'
```
EmEhPa: why not use ref and computed?

src/front_end/src/components/SearchFiltersCard.vue
EmEhPa, line 127: Why is so much css code removed

src/front_end/src/views/SearchView.vue
```diff
@@ -85,8 +95,39 @@ const closePreview = () => {
 }
 
 /* Handle changes to search filters (currently just logs the change) */
-const handleFilterChange = (filter) => {
-  console.log('Filter changed:', filter)
+const handleFilterChange = (filters) => {
```
EmEhPa, line 99: This is already kind of done in compostables/useSearchMatches,js,

EmEhPa (changes requested):

src/front_end/src/components/SearchFiltersCard.vue
muaathaw: reactive fits better here since we're managing a multi-key object that updates together. watch handles the parent sync. Works as intended.

src/front_end/src/components/SearchFiltersCard.vue
muaathaw: The removed CSS was unused — the classes were never referenced in the template. Keeping dead code would be bad practice. :)

src/front_end/src/views/SearchView.vue
muaathaw: Good catch, already refactored to use resolveDocumentType from the composable. The manual extension checks are removed.I'll commit the changes and test tomorrow. If it doesn't work as expected we can revisit, otherwise the duplicate logic is already worked.

EmEhPa:
Optimaztion of [composables/useSearchMetadata.js] so its a three dict later.

PR #261, opened 2026-03-27
Refactor authentication logic and implement automatic token refresh

muaathaw:
## Description
This PR refactors the authentication flow and avoids duplicate logic between `AuthCallbackView.vue` and `auth.js`.

The token handling logic is now centralized in `auth.js`, making the code more maintainable and easier to extend.

## Changes
- Centralized all token-related logic in `auth.js`
- Added reusable `requestToken()` function for:
  - Authorization code exchange
  - Refresh token flow
- Implemented `getValidToken()` to:
  - Automatically check token expiration
  - Refresh the token if it is about to expire
- Standardized token storage (`expires_at`, `access_token`, etc.)
- Improved error handling and cleanup logic

### DIFF
--- a/src/front_end/src/utils/auth.js
+++ b/src/front_end/src/utils/auth.js
@@ -90,7 +90,7 @@ export async function refreshTokenRequest(refreshToken) {
 
     // If refresh fails → force logout
     clearAuth()
-    window.location.href = '/' // send user to login page 
+    window.location.href = '/' // send user to login page
 
     return null
   }
--- a/src/front_end/src/utils/auth.js
+++ b/src/front_end/src/utils/auth.js
@@ -35,6 +35,10 @@ function clearAuth() {
   sessionStorage.removeItem('refresh_token')
   sessionStorage.removeItem('id_token')
   sessionStorage.removeItem('expires_at')
+  sessionStorage.removeItem('pkce_verifier')
+  sessionStorage.removeItem('oidc_state')
+
+  localStorage.setItem('logout_event', Date.now().toString())
 }
 
 /* Generic Token Request */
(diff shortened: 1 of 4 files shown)


src/front_end/src/utils/auth.js
```diff
@@ -5,8 +5,113 @@
...
+    grant_type: 'authorization_code',
+    client_id: CLIENT_ID,
+    code,
+    redirect_uri: redirectUri,
+    code_verifier: verifier
+  })
+}
+
+/*Refresh access token using refresh token */
+export async function refreshTokenRequest(refreshToken) {
```
EmEhPa, line 79: where is this function called?

src/front_end/src/views/AuthCallbackView.vue
```diff
@@ -5,12 +5,9 @@
  */
 import { onMounted, ref } from 'vue'
 import { useRoute, useRouter } from 'vue-router'
+import { exchangeCodeForToken } from '../utils/auth'
```
EmEhPa: We use @ instead of .. everywhere else.

EmEhPa (changes requested):
This doesnt refresh from what I and Oskar can see.

src/front_end/src/utils/auth.js
```diff
@@ -5,8 +5,113 @@
...
+      refresh_token: refreshToken
+    })
+
+    return data.access_token
+  } catch (error) {
+    console.error('Refresh failed:', error)
+
+    // If refresh fails → force logout
+    clearAuth()
+    window.location.href = '/login'
```
EmEhPa: change to only '/'
As far as i know we dont have path '/login'

EmEhPa (changes requested):

Issue #234, opened 2026-03-23
Construct unittest framework


jaen22BTH:
Construct unit tests and point to it in automated linters.

PR #264, opened 2026-03-29
Addition of automatically running unittests for Gitlabs connector (re…

jaen22BTH:
Addition of automatically running unittests for Gitlabs connector (ref #234)


AlexanderJarnstromBTH (approved):
crazy

PR #271, opened 2026-03-31
Feature/search engine/unit tests

AlexanderJarnstromBTH:


src/search_engine/src/se_api/services/connector.py
```diff
@@ -189,3 +186,25 @@ def _files_to_index(self) -> str | None:
         self.subdata = subdata
 
         return file_url
+
+    def _get_file_pointers(self) -> Any | None:
+        return get(
```
jaen22BTH: Do we want error handling here, considering timeout is set (or non JSON parseble object is returned), I believe this will throw JsonDecodeError, maybe catch that and throw dmis_warning? Same is true for all get methods below.

AlexanderJarnstromBTH: the error handling is currently in the parent function, though it could/should be moved down.

src/search_engine/src/se_api/services/connector.py
```diff
@@ -189,3 +135,77 @@ def _files_to_index(self) -> str | None:
...
+            dms_warning(f"Failed to connect, url: {self.url_files_to_index}.")
+        except exceptions.HTTPError:
+            dms_warning(f"Invalid HTTP response, url: {self.url_files_to_index}.")
+        except exceptions.Timeout:
+            dms_warning(f"Request timed out, url: {self.url_files_to_index}")
+        except exceptions.JSONDecodeError:
+            dms_warning(f"Failed to parse JSON, url: {self.url_files_to_index}.")
+        except exceptions.RequestException:
+            dms_warning(f"Something went wrong, url: {self.url_files}.")
+        return None
```
mamz22, line 211: The methods R138-R211 have lots of duplicated request/exception handling logic. This could potentially be refactored into a single shared helper

src/search_engine/src/se_api/services/connector.py
```diff
@@ -189,3 +135,77 @@ def _files_to_index(self) -> str | None:
...
+
+    def _get_file_from_pointer(self, pointer: str) -> Any | None:
+        """Get file from pointer"""
+        try:
+            return get(
+                self.url_file,
+                params=[("file_pointer", pointer), ("include_content", False)] if self.subdata is not None else None,
```
mamz22: Might be missing some context here but as its written this makes the entire params payload conditional based on self.subdata, so file_pointer is left out when self.subdata is None. If file_pointer is required for this endpoint, the params construction should be changed so that file_pointer is always included and only subdata is conditional

AlexanderJarnstromBTH: good catch, has probably been left there when copying from a request actually having a conditional subdata option.

src/search_engine/src/se_api/services/connector.py
AlexanderJarnstromBTH: grabbing a generic exception and then handling it in a function didnt make tox happy.

jaen22BTH (approved):

Issue #269, opened 2026-03-31
Move over to Multi-language model


Sjomullen:
- [x] Find a new model and get it up and running on one of the GPU servers
- [ ] Verify that the new model works similarly to the original English only model
- [ ] Remove the old model

PR #273, opened 2026-04-01
#269 Feature/classtune

Sjomullen:
#269 Made some changes to label triggers. Also added a loop to escalate classification if the values are too close to each other, usual cause seems to be that the zero-shot model is unsure of what classification to put. There for it is safe to escalate the classification.


src/stochastic_analyzer/gateway/services/classifier.py
```diff
@@ -34,13 +41,28 @@ def _build_inputs(items: list[InputItem], max_chars: int) -> list[list[str]]:
 
 def _resolve_labels(items: list[InputItem], all_scores: list[float]) -> list[ClassificationResult]:
     """Map entailment scores back to classification labels per document."""
+    escalation_threshold = 0.02
```
AlexanderJarnstromBTH: maybe dont want this hard coded, env var?

Sjomullen: very true very true, will fix asap

jaen22BTH (changes requested):
Archive this, and replace this with #279?

PR #274, opened 2026-04-01
Feature/main api/aisummary

EmEhPa:
I belive the commit messages explains a lot but a whole summary:

Classifiers fixed, you can also filter on them
Summarization is fixed
Filter on file type works and will be refined next week.

The linters problem: pls comment on them and Marko will look into that this afternoon :)

### DIFF
--- a/src/api/src/dmis_api/api.py
+++ b/src/api/src/dmis_api/api.py
@@ -138,12 +138,19 @@ class API:
         # return JSONResponse(status_code=200, content=data)
 
         # extract unique pointers from search metadata
-        results = search_data.get("results", [])
+        if isinstance(search_data, list):
+            results = search_data
+        elif isinstance(search_data, dict):
+            results = search_data.get("results", [])
+        else:
+            dms_warning("Search API returned unexpected JSON shape.")
+            return JSONResponse(status_code=502, content="")
+
         if not isinstance(results, list):
             dms_warning("Search API response missing valid 'results' list.")
             return JSONResponse(status_code=502, content="")
-        
-        unique_pointers = list[str] = []
+                
+        unique_pointers: list[str] = []
         for entry in results:
             if not isinstance(entry, dict):
                 continue
@@ -160,7 +167,7 @@ class API:
             return JSONResponse(
                 status_code=200,
                 content={
-                    "results"; [],
+                    "results": [],
                     "query": query,
                     "detail": ""
                 },
@@ -180,7 +187,7 @@ class API:
             )
         except requests.RequestException as exc:
             dms_warning(f"Query service request failed: {exc}")
-            return JSONResponse(status_code=502, content=""),
+            return JSONResponse(status_code=502, content="")
         
         if not query_response.ok:
             dms_warning(f"Query service returned status code {query_response.status_code}.")
--- a/src/api/src/dmis_api/api.py
+++ b/src/api/src/dmis_api/api.py
@@ -52,7 +52,7 @@ class API:
         search_api_url = os.getenv("DMIS_SEARCH_API_URL")
         if not isinstance(search_api_url, str) or not search_api_url:
             dms_warning("DMIS_SEARCH_API_URL is not set.")
-            return JSONResponse(status_code=500, content="D")
+            return JSONResponse(status_code=500, content="")
 
         try:
             response = requests.get(
@@ -62,20 +62,50 @@ class API:
             )
         except requests.RequestException as exc:
             dms_warning(f"Search request to upstream API failed: {exc}")
-            return JSONResponse(status_code=502, content="A")
+            return JSONResponse(status_code=502, content="")
 
         if not response.ok:
             dms_warning(f"Upstream search API returned status code {response.status_code}.")
-            return JSONResponse(status_code=502, content="B")
+            return JSONResponse(status_code=502, content="")
 
         try:
             data = response.json()
         except requests.JSONDecodeError as exc:
             dms_warning(f"Upstream search API returned invalid JSON: {exc}")
-            return JSONResponse(status_code=502, content="C")
+            return JSONResponse(status_code=502, content="")
 
         return JSONResponse(status_code=200, content=data)
+    
+    @staticmethod
+    @app.get("/summary", status_code=200)
+    async def summary(file_pointer: str) -> JSONResponse:
+        """Forward summary request to upstream summary API."""
+        dmis_summary_url = os.getenv("DMIS_SUMMARY_API_URL")
+        if not isinstance(dmis_summary_url, str) or not dmis_summary_url:
+            dms_warning("DMIS_SUMMARY_API_URL is not set.")
+            return JSONResponse(status_code=500, content="")
 
+        try:
+            response = requests.get(
+                f"{dmis_summary_url.rstrip('/')}/summarize",
+                params={"unique_pointer": file_pointer},
+                timeout=120,
+            )
+        except requests.RequestException as exc:
+            dms_warning(f"Summary request to upstream API failed: {exc}")
+            return JSONResponse(status_code=502, content="")
+
+        if not response.ok:
+            dms_warning(f"Upstream summary API returned status code {response.status_code}.")
+            return JSONResponse(status_code=502, content="")
+
+        try:
+            data = response.json()
+        except requests.JSONDecodeError as exc:
+            dms_warning(f"Upstream summary API returned invalid JSON: {exc}")
+            return JSONResponse(status_code=502, content="")
+
+        return JSONResponse(status_code=200, content=data)
 
 def run() -> None:
     """Initiate FastAPI using Uvicorn."""
@@ -87,9 +117,9 @@ def run() -> None:
     if args.dev:
         api.log_level = "debug"
 
-    port = os.getenv("API_PORT", "8000")
+    port = os.getenv("API_PORT")
     if port is None or not port.isdigit():
-        dms_error("Port for DMIS API not set  as digit in local environment, please export API_PORT.")
+        dms_error("Port for DMIS API not set as digit in local environment, please export API_PORT.")
         return
 
     uvicorn.run(
--- a/src/api/src/dmis_api/api.py
+++ b/src/api/src/dmis_api/api.py
@@ -13,7 +13,7 @@ from fastapi.encoders import jsonable_encoder
 from fastapi.exceptions import RequestValidationError
 from fastapi.responses import JSONResponse
 
-from dmis_logger import dms_warning, dms_error
+from dmis_logger import dms_info, dms_warning, dms_error
 from fastapi import Query
 
 # Testing
@@ -21,29 +21,65 @@ from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import PlainTextResponse
 
 
-
 class API:
     """Management class for main API."""
 
-
     app: FastAPI = FastAPI()
 
-    app.add_middleware(
-        CORSMiddleware,
-        allow_origins=["http://localhost:8080"],  # frontend URL
-        allow_credentials=True,
-        allow_methods=["*"],
-        allow_headers=["*"],
-        )
-    
     log_level: str | None = None
+    bind_address: str
+    port: int
 
     def __init__(self) -> None:
         """Constructor."""
+        bind_address = os.environ.get("API_BIND_ADDRESS")
+        port = os.environ.get("API_PORT")
+        allowed_origins = os.environ.get("API_ALLOWED_ORIGINS")
+
+        if bind_address is None:
+            dms_error("API_BIND_ADDRESS is not defined.")
+            return
+        if port is None:
+            dms_error("API_PORT is not defined.")
+            return
+        if not port.isdigit():
+            dms_error("API_PORT expected integer.")
+            return
+        if int(port) <= 0 or int(port) >= 65535:
+            dms_error("API_PORT should be between 0 and 65535.")
+            return
+
+        parser = argparse.ArgumentParser()
+        _ = parser.add_argument("--dev", action="store_true")
+        args = parser.parse_args()
+
+        if args.dev:
+            self.log_level = "debug"
+
+        self.bind_address = bind_address
+        self.port = int(port)
+
         self.app.add_exception_handler(
             RequestValidationError,
             self.validation_exception_handler,
         )
+
+        if allowed_origins is not None:
+            dms_info(f"Allowing origins: {allowed_origins.split(',')}")
+            self.app.add_middleware(
+                CORSMiddleware,
+                allow_origins=allowed_origins.split(","),
+                allow_credentials=True,
+                allow_methods=["*"],
+                allow_headers=["*"],
+            )
+
+        self.app.add_api_route("/search", self.search, methods=["GET"])
+        self.app.add_api_route("/summary", self.summary, methods=["GET"])
+
+    def start(self) -> None:
+        """Start API"""
+        uvicorn.run(self.app, host=self.bind_address, port=self.port, log_level=self.log_level)
 
     async def validation_exception_handler(self, _: Request, exc: Exception) -> JSONResponse:
         """Overwrite FastAPI exception handler."""
@@ -61,9 +97,7 @@ class API:
 
         return JSONResponse(status_code=422, content=content)
 
-    @staticmethod
-    @app.get("/search", status_code=200)
-    async def search(query = Query(..., min_length=1, max_length=200)) -> JSONResponse:
+    async def search(self, query=Query(..., min_length=1, max_length=200)) -> JSONResponse:
         """Forward search request to upstream search API."""
         search_api_url = os.getenv("DMIS_SEARCH_API_URL")
         if not isinstance(search_api_url, str) or not search_api_url:
@@ -91,11 +125,9 @@ class API:
             return JSONResponse(status_code=502, content="")
 
         return JSONResponse(status_code=200, content=data)
-    
+
     # Currently works with the example form query, we will change the payload to have unique pointer once that has been fixed.
-    @staticmethod
-    @app.get("/summary")
-    async def summary(file_pointer: str = Query(..., min_length=1, max_length=500)) :
+    async def summary(self, file_pointer: str = Query(..., min_length=1, max_length=500)):
         """Forward summary request to upstream summary API and return plain text."""
         dmis_summary_url = os.getenv("DMIS_SUMMARY_API_URL")
         if not dmis_summary_url:
@@ -110,7 +142,6 @@ class API:
                     "name": "Incident_Report_1.txt",
                     "author": "IT Admin",
                 },
-
             },
             {
                 "content": "Power was fully restored to the main grid by 4 AM. No data loss was reported.",
@@ -118,16 +149,11 @@ class API:
                     "name": "Incident_Report_2.txt",
                     "author": "IT Admin",
                 },
-
-            }
+            },
         ]
 
         try:
-            response = requests.post(
-                f"{dmis_summary_url.rstrip('/')}/summarize",
-                json=payload,
-                timeout=120
-            )
+            response = requests.post(f"{dmis_summary_url.rstrip('/')}/summarize", json=payload, timeout=120)
             response.raise_for_status()
         except requests.RequestException as exc:
             dms_warning(f"Summary request to upstream API failed: {exc}")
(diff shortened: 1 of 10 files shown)


src/api/src/dmis_api/api.py
```diff
@@ -45,56 +99,211 @@ async def validation_exception_handler(self, _: Request, exc: Exception) -> JSON
...
+        query_api_url = os.getenv("DMIS_QUERY_API_URL")
+
+        if not search_api_url:
             dms_warning("DMIS_SEARCH_API_URL is not set.")
-            return JSONResponse(status_code=500, content="D")
+            return self._error_response(500)
+
+        if not query_api_url:
+            dms_warning("DMIS_QUERY_API_URL is not set.")
+            return self._error_response(500)
```
AlexanderJarnstromBTH: should be moved to the constructor, with the other env variables

src/api/src/dmis_api/api.py
```diff
@@ -45,56 +99,211 @@ async def validation_exception_handler(self, _: Request, exc: Exception) -> JSON
...
+            dms_warning("Search request received empty query.")
+            return self._error_response(422)
 
         try:
-            response = requests.get(
+            search_response = requests.get(
                 f"{search_api_url.rstrip('/')}/search",
                 params={"q": query},
-                timeout=120,
+                timeout=30,
```
AlexanderJarnstromBTH: should probably be ~120. 30 sec will fail on a reindex

src/api/src/dmis_api/api.py
```diff
@@ -8,27 +8,81 @@
...
+
+        self.app.add_api_route("/search", self.search, methods=["GET"])
+        self.app.add_api_route("/summary", self.summary, methods=["POST"])
+
+    def start(self) -> None:
+        """Start API"""
+        uvicorn.run(self.app, host=self.bind_address, port=self.port, log_level=self.log_level)
+
+    @staticmethod
+    def _error_response(status_code: int) -> JSONResponse:
```
AlexanderJarnstromBTH: Could use:
```python
async def validation_exception_handler(self, _: Request, exc: Exception) -> JSONResponse:
        """Overwrite FastAPI exception handeler."""
        errors: dict
        if isinstance(exc, RequestValidationError):
            errors = {"detail": exc.errors(), "body": exc.body}
        else:
            errors = {"detail": str(exc)}
        content: str | dict
        if self.log_level == "debug":
            content = jsonable_encoder(errors)
        else:
            content = "ERROR"
        return JSONResponse(status_code=422, content=content)
```

src/api/src/dmis_api/api.py
```diff
@@ -45,56 +99,211 @@ async def validation_exception_handler(self, _: Request, exc: Exception) -> JSON
...
-            return JSONResponse(status_code=502, content="B")
+            return self._error_response(502)
 
         try:
-            data = response.json()
+            search_data = search_response.json()
         except requests.JSONDecodeError as exc:
             dms_warning(f"Upstream search API returned invalid JSON: {exc}")
-            return JSONResponse(status_code=502, content="C")
+            return self._error_response(502)
```
AlexanderJarnstromBTH: could call and let fastAPI handle it (will remove tox complaints about too many returns):
```python
raise HTTPException(status_code=502)
```
apply this comment to all instances of this

src/api/src/dmis_api/api.py
```diff
@@ -45,56 +99,211 @@ async def validation_exception_handler(self, _: Request, exc: Exception) -> JSON
 
         return JSONResponse(status_code=422, content=content)
 
-    @staticmethod
-    @app.get("/search", status_code=200)
-    async def search(query: str) -> JSONResponse:
-        """Forward search request to upstream search API."""
+    async def search(self, query: str = Query(..., min_length=1, max_length=200)) -> JSONResponse:
```
AlexanderJarnstromBTH, line 81: could probably split this into several service functions

src/api/src/dmis_api/api.py
```diff
@@ -45,56 +99,211 @@ async def validation_exception_handler(self, _: Request, exc: Exception) -> JSON
...
+
+        query = query.strip()
+        if not query:
+            dms_warning("Search request received empty query.")
+            return self._error_response(422)
 
         try:
-            response = requests.get(
+            search_response = requests.get(
                 f"{search_api_url.rstrip('/')}/search",
```
AlexanderJarnstromBTH: strip of the `/` in the constructor when you grab the environment variable, removes the need for repetitive calls to `.rstrip()`.
Apply to all other instances as well

src/api/src/dmis_api/api.py
mamz22: Resolved

src/api/src/dmis_api/api.py
```diff
@@ -45,56 +106,212 @@ async def validation_exception_handler(self, _: Request, exc: Exception) -> JSON
...
+        except requests.RequestException as exc:
+            response_text = exc.response.text if exc.response is not None else ""
+            dms_warning(f"Classification request failed: {exc}. " f"Response body: {response_text}")
+            raise HTTPException(status_code=502) from exc
+
+        try:
+            classification_data = response.json()
+        except requests.JSONDecodeError as exc:
+            dms_warning(f"Classification service returned invalid JSON: {exc}")
+            raise HTTPException(status_code=502) from exc
```
AlexanderJarnstromBTH: can merge these two:
```python
classification_data = requests.post(
    f"{self.query_api_url}/classify",
    json={"pointers": pointers},
    timeout=30,
).json()
```
just remember to catch both potential exceptions

src/api/src/dmis_api/api.py
```diff
@@ -45,56 +106,212 @@ async def validation_exception_handler(self, _: Request, exc: Exception) -> JSON
...
+
+        if isinstance(classification_data, list):
+            classification_items = [item for item in classification_data if isinstance(item, dict)]
+        elif isinstance(classification_data, dict):
+            classification_items = [classification_data]
```
AlexanderJarnstromBTH: does the query `/classify` return anything other then a list or None?

src/api/src/dmis_api/api.py
```diff
@@ -8,27 +8,88 @@
...
+            raise RuntimeError("Startup failed.")
+        if int(port) <= 0 or int(port) >= 65535:
+            dms_error("API_PORT should be between 0 and 65535.")
+            raise RuntimeError("Startup failed.")
+        if not search_api_url:
+            dms_error("DMIS_SEARCH_API_URL is not set.")
+            raise RuntimeError("Startup failed.")
+        if not query_api_url:
+            dms_error("DMIS_QUERY_API_URL is not set.")
+            raise RuntimeError("Startup failed.")
```
AlexanderJarnstromBTH: `dms_error()` calls `os._exit()`. no need to raise exception

src/api/src/dmis_api/api.py
```diff
@@ -45,56 +106,212 @@ async def validation_exception_handler(self, _: Request, exc: Exception) -> JSON
...
-            return JSONResponse(status_code=502, content="B")
+            raise HTTPException(status_code=502) from exc
 
         try:
-            data = response.json()
+            search_data = response.json()
         except requests.JSONDecodeError as exc:
             dms_warning(f"Upstream search API returned invalid JSON: {exc}")
-            return JSONResponse(status_code=502, content="C")
+            raise HTTPException(status_code=502) from exc
```
AlexanderJarnstromBTH, line 101: can merge these two:
```python
search_data = requests.get(
    f"{self.search_api_url}/search",
    params={"q": query},
    timeout=120,
).json()
```
just remember to catch both potential exceptions

src/api/src/dmis_api/api.py
```diff
@@ -45,56 +106,212 @@ async def validation_exception_handler(self, _: Request, exc: Exception) -> JSON
...
+            results = search_data
+        elif isinstance(search_data, dict):
+            results = search_data.get("results", [])
+        else:
+            dms_warning("Search API returned unexpected JSON shape.")
+            raise HTTPException(status_code=502)
+
+        if not isinstance(results, list):
+            dms_warning("Search API response missing valid 'results' list.")
+            raise HTTPException(status_code=502)
```
AlexanderJarnstromBTH: could probably get away with just checking if the response is a list or not

src/api/src/dmis_api/api.py
```diff
@@ -45,56 +106,212 @@ async def validation_exception_handler(self, _: Request, exc: Exception) -> JSON
...
+        """Forward search request to upstream search API, enrich results with classification, and return results."""
+        query = query.strip()
+        if not query:
+            dms_warning("Search request received empty query.")
+            raise HTTPException(status_code=422)
+
+        results = self.fetch_search_results(query)
+
+        unique_pointers: list[str] = []
+        seen_pointers: set[str] = set()
```
AlexanderJarnstromBTH: whats the purpose of seen_pointers.

src/api/src/dmis_api/api.py
```diff
@@ -45,56 +106,212 @@ async def validation_exception_handler(self, _: Request, exc: Exception) -> JSON
...
+                    if entry_name:
+                        classification_value = classification_map_by_name.get(entry_name)
+
+            if isinstance(classification_value, dict):
+                security_class = classification_value.get("Security-class")
+                if isinstance(security_class, str) and security_class.strip():
+                    entry["security_class"] = security_class.strip()
+                else:
+                    entry["security_class"] = None
+            else:
```
FilipDar: Double Else statments could be cleaner

src/api/src/dmis_api/api.py
```diff
@@ -45,56 +106,212 @@ async def validation_exception_handler(self, _: Request, exc: Exception) -> JSON
...
+                seen_pointers.add(pointer)
+
+        if not unique_pointers:
+            return JSONResponse(
+                status_code=200,
+                content={
+                    "results": [],
+                    "query": query,
+                    "detail": "",
+                },
```
AlexanderJarnstromBTH: could simply return the empty list, need to change the return type to list instead of JSONResponse.

src/api/src/dmis_api/api.py
```diff
@@ -45,56 +106,212 @@ async def validation_exception_handler(self, _: Request, exc: Exception) -> JSON
...
+            classification_map_by_name,
+        )
+
+        return JSONResponse(
+            status_code=200,
+            content={
+                "results": results,
+                "query": query,
+            },
+        )
```
AlexanderJarnstromBTH, line 113: could simply return the empty list, need to change the return type to list instead of JSONResponse.

src/api/src/dmis_api/api.py
```diff
@@ -45,56 +106,212 @@ async def validation_exception_handler(self, _: Request, exc: Exception) -> JSON
...
+            response = requests.post(
+                f"{self.query_api_url}/summarize",
+                json=payload,
+                timeout=100,
+            )
+            response.raise_for_status()
+        except requests.RequestException as exc:
+            response_text = ""
+            if hasattr(exc, "response") and exc.response is not None:
+                response_text = exc.response.text
```
AlexanderJarnstromBTH, line 142: what is happening here?

src/api/src/dmis_api/api.py
mamz22: the purpose of seen_pointers is to deduplicate unique_pointer values before calling fetch_classification_maps, while unique_pointers preserves the original order

src/api/src/dmis_api/api.py
```diff
@@ -8,27 +8,88 @@
...
+
     log_level: str | None = None
+    bind_address: str
+    port: int
+    search_api_url: str
+    query_api_url: str
 
     def __init__(self) -> None:
         """Constructor."""
+        bind_address = os.environ.get("API_BIND_ADDRESS")
```
jaen22BTH: Could we maybe move some of the functionality not related to this class closer to the entrypoint, before constructor is called.

src/api/src/dmis_api/api.py
```diff
@@ -8,27 +8,88 @@
...
+        if bind_address is None:
+            dms_error("API_BIND_ADDRESS is not defined.")
+            raise RuntimeError("Startup failed.")
+        if port is None:
+            dms_error("API_PORT is not defined.")
+            raise RuntimeError("Startup failed.")
+        if not port.isdigit():
+            dms_error("API_PORT expected integer.")
+            raise RuntimeError("Startup failed.")
+        if int(port) <= 0 or int(port) >= 65535:
```
jaen22BTH: Unnecessary to cast to cast this to int multiple times (also done on line 66)

src/api/src/dmis_api/api.py
```diff
@@ -8,27 +8,88 @@
...
+            dms_error("API_PORT should be between 0 and 65535.")
+            raise RuntimeError("Startup failed.")
+        if not search_api_url:
```
jaen22BTH: Maybe would clutter the code a bit to just do this and below in two lines instead.

src/api/src/dmis_api/api.py
```diff
@@ -8,27 +8,88 @@
...
+        parser = argparse.ArgumentParser()
+        _ = parser.add_argument("--dev", action="store_true")
+        args = parser.parse_args()
+
+        if args.dev:
+            self.log_level = "debug"
+
+        self.bind_address = bind_address
+        self.port = int(port)
+        self.search_api_url = search_api_url.rstrip("/")
```
jaen22BTH, line 45: Why read this into local var in method, just to then read it into class var?

src/api/src/dmis_api/api.py
```diff
@@ -8,27 +8,88 @@
...
+
         self.app.add_exception_handler(
             RequestValidationError,
             self.validation_exception_handler,
         )
 
+        if allowed_origins is not None:
+            origins = [origin.strip() for origin in allowed_origins.split(",") if origin.strip()]
+            dms_info(f"Allowing origins: {origins}")
+            self.app.add_middleware(
```
jaen22BTH: Is the thinking that we will have a reverse proxy in front of this?

src/api/src/dmis_api/api.py
```diff
@@ -45,56 +106,212 @@ async def validation_exception_handler(self, _: Request, exc: Exception) -> JSON
...
-    async def search(query: str) -> JSONResponse:
-        """Forward search request to upstream search API."""
-        search_api_url = os.getenv("DMIS_SEARCH_API_URL")
-        if not isinstance(search_api_url, str) or not search_api_url:
-            dms_warning("DMIS_SEARCH_API_URL is not set.")
-            return JSONResponse(status_code=500, content="D")
-
+    def fetch_search_results(self, query: str) -> list[Any]:
+        """Fetch search results from upstream search API."""
         try:
```
jaen22BTH, line 88: Could we just merge this into a large try catch statement instead?

src/api/src/dmis_api/api.py
```diff
@@ -45,56 +106,212 @@ async def validation_exception_handler(self, _: Request, exc: Exception) -> JSON
...
+                status_code=200,
+                content={
+                    "results": [],
+                    "query": query,
+                    "detail": "",
+                },
+            )
+
+        classification_map_by_pointer, classification_map_by_name = self.fetch_classification_maps(unique_pointers)
+        self.apply_security_classes(
```
jaen22BTH: We really want to minimize unnecessary logic and parsing, to make it possible to maintain the application. If all the mapping in this method really is needed, it should probably be communicated with the query side to return a structure to allow for this.

src/front_end/src/components/SearchFiltersCard.vue
```diff
@@ -1,55 +1,72 @@
 <script setup>
-import { computed, ref } from 'vue'
+import { ref, computed } from 'vue'
 import { Grid2X2, FileText, Shield } from 'lucide-vue-next'
 
-const sourceFilters = ['GitHub', 'GitLab', 'Network File System'] // Add more sources needed if possible
+const sourceFilters = ['GitHub', 'GitLab', 'Network File System']
```
jaen22BTH, line 5: Should this really exist in the frontend? We need to make the application dynamic, meaning this probably should be supplied from lower levels (should probably be implemented at connector level).

src/api/src/dmis_api/api.py
```diff
@@ -8,27 +8,88 @@
 
 import requests
```
jaen22BTH, line 9: I might be wrong here, but I think we generally want to avoid implementing, parsing, mapping or logic in the API, it makes the application extremely difficult to maintain if it isn't structurally clear where logic is implemented. 

From that perspective, a 'warning flag' might be if the API file is more than ~150 lines of code. Has the system component responsibility been carefully considered at each step here?

jaen22BTH (changes requested):

src/api/src/dmis_api/api.py
mamz22: I understand the concern. My interpretation was that this was a part of the main API’s responsibility, but I agree the API layer itself is currently be doing too much. I could move the upstream fetch/parsing and logic into helper or service functions so the endpoint stays thinner.

I am not sure I fully understand your last sentence, are you suggesting that some of this is outside the scope of the main API component or that it should remain in the component but be structured more clearly across separate layers?

src/api/src/dmis_api/api.py
mamz22: @kebabohvo added this and if I remember it correctly it has something to do with testing the local API testing work together with the frontend. So my understanding is that it may be needed for local development, but I am not sure whetever it is also intended to stay in production?

src/api/src/dmis_api/api.py
mamz22: No it doesn't

src/api/src/dmis_api/api.py
mamz22: Okay, I could either move that into a separate service/helper or discuss with the query side whether they can return a structure that makes this mapping unnecessary here, what is prefered?

src/api/src/dmis_api/api.py
mamz22: If the requests exception contains a response, we read its body and include it in the warning log before translating the failure into a 502 for our API caller. 

Will probably change that code snippet to this tomorrow
response_text = exc.response.text if exc.response is not None else ""
dms_warning(
    f"Summary request to upstream API failed: {exc}. Response body: {response_text}"
)
raise HTTPException(status_code=502) from exc

src/front_end/src/components/SearchFiltersCard.vue
EmEhPa: Not really but for now we only have GitLab and we wanted some kind of filter for the visualization, which has been there from the beginning. It's something we plan to work on later, like the beginning of sprint 4. The plan has never been to leave this hardcoded but make it mor dynamic in the furure..

src/api/src/dmis_api/api.py
AlexanderJarnstromBTH: to handle potential input errors without the need to have `<type> | None` for the class level variables. Removing the need to perform unnecessary type checking in the functions using them.

src/api/src/dmis_api/api.py
AlexanderJarnstromBTH: so two if-statements, if i understand correctly

src/api/src/dmis_api/api.py
mamz22: So I understand this correctly, do you mean moving startup/bootstrap concerns such as argparse and log-level handling closer to run(), or do you also want env/config loading and validation to be moved out of the constructor?

src/api/src/dmis_api/api.py
mamz22: This function might get deleted because the responsibility of fetching classification might be moved to the Search Engine

src/api/src/dmis_api/api.py
mamz22: This function might get deleted because the responsibility of fetching classification might be moved to the Search Engine

src/api/src/dmis_api/api.py
mamz22: I have talked with Alex about the responsibility of fetching the classification data and appending it into the metadata of each file, he might come up with a solution through the Search Engine side. This would result in that the fetching of classification metadata would be included automatically from the main api's point of view when calling the search engine endpoint with a query.

src/front_end/src/components/SearchFiltersCard.vue
jaen22BTH: Sure, but, would it not be better to keep hardcoded assumtions out of the frontend even if concidered temporary (espessially in this case where it only would require one line of code in the connector layer to solve). Or what is the point of building a 'fileter' for some abstract dev environment. Idk, maybe we just need to implement a mechanism to document these things, and it won't be a problem :smile: ? (My bad, should have asked this in the previous commit).

EmEhPa: we do a new branch for this to fix. Promise

src/api/src/dmis_api/api.py
mamz22: Is gonna get removed since local api testing between front end and api should not happen

PR #279, opened 2026-04-01
#269 merged with other Filips code and ran some tests on it as well

Sjomullen:
* Merged with other Filips code and get the same results as before

### DIFF
--- a/src/stochastic_analyzer/gateway/services/classifier.py
+++ b/src/stochastic_analyzer/gateway/services/classifier.py
@@ -11,10 +11,17 @@ from gateway.schemas import InputItem, ClassificationResult
 LABELS = ["Public", "Internal", "Sensitive", "Confidential"]
 
 LABEL_TRIGGERS = [
-    "This information is common knowledge and intended for the general public.",
-    "This is standard corporate communication for regular employees.",
-    "This document contains restricted technical or operational data like passwords or keys.",
-    "This is a strictly secret document containing high-level CEO strategy, mergers, or financial projections.",
+    # Public
+    "This is general information intended for the public, such as manuals, public announcements, or open event invitations.",
+    # Internal
+    "This is internal company information meant only for employees, such as sales targets, project plans, "
+    "team updates, system migrations, or internal process changes.",
+    # Sensitive
+    "This document contains sensitive employee or operational data such as performance reviews, "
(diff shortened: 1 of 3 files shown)


src/stochastic_analyzer/gateway/services/classifier.py
```diff
@@ -32,27 +39,48 @@ def _build_inputs(items: list[InputItem], max_chars: int) -> list[list[str]]:
     return inputs
 
 
-def _resolve_labels(items: list[InputItem], all_scores: list[float]) -> list[ClassificationResult]:
+def _escalate(doc_scores: list[float], best_index: int, escalation_threshold: float) -> int:
```
FilipDar: Maybe have a default failsafe treshold variable to fall back on if no env specifics are requested?
The function should probably run without hard specified but leave it as a tuning parameter perhaps

src/stochastic_analyzer/gateway/services/classifier.py
```diff
@@ -32,27 +39,48 @@ def _build_inputs(items: list[InputItem], max_chars: int) -> list[list[str]]:
...
+    label_rank = {"Public": 0, "Internal": 1, "Sensitive": 2, "Confidential": 3}
+    best_score = doc_scores[best_index]
+
+    for i, score in enumerate(doc_scores):
+        is_higher_rank = label_rank[LABELS[i]] > label_rank[LABELS[best_index]]
+        is_within_threshold = (best_score - score) < escalation_threshold
+
+        if is_higher_rank and is_within_threshold:
+            best_index = i
+            best_score = score
```
FilipDar, line 58: Should this stay as the original winners score? Updating it here means each escalation makes the next one easier to trigger. causing like some cascading effect?

Sjomullen: if the score between the next two classifications are within the set value there will be a cascading effect into an even higher security classification. Not sure how likely it will be that the values will be that close. I think it happened in one case out of 42 or 45 test cases. Not sure what the specific case contained though

src/stochastic_analyzer/gateway/services/classifier.py
Sjomullen: Have added that now, will run tests after easter and when that is done push the updated code

FilipDar (approved):

PR #281, opened 2026-04-01
Feature/smb

orca365:


jaen22BTH (changes requested):
Why is the entire gitlab and storage infra moved? This will cause large problems in many system parts. Please no not change anything related to these files in this commit, should be done in separate commit if changes are wanted.

PR #284, opened 2026-04-06
Add GitHub connector baseline and OAuth support (ref #58)

### LINKED ISSUE #58: Integrate with GitHub
- As an IT/system integrator, I want to connect the system to GitHub repositories, so that code documentation and wiki content becomes searchable alongside other documents.
- Acceptance Criteria:
  - System successfully authenticates with GitHub via official API
  - Documents from both platforms are ingested and indexed
  - User permissions from source systems are respected

bjpo78:
## Summary

Adds the GitHub connector under \`src/connectors/githubs\`, aligned with the existing GitLab connector's API contract (\`/files\`, \`/file\`, \`/files_to_index\`) and payload shape for the search/indexer integration.

Authentication is per-user, not static: each request must carry an \`X-GitHub-Token\` header with the user's own token. A missing header returns an empty result immediately — no GitHub API calls are made. There is no service-level fallback token.

## What's included

- **Connector implementation**: \`github_collector_api.py\` (FastAPI) and \`interfacer_github.py\` (GitHub API client, file pointers, bulk indexing via archive).
- **Per-user auth passthrough**: \`connector.py\` in the search engine forwards source-system tokens as named headers via \`_TOKEN_HEADER_MAP\`. Adding a future system (Confluence, Google Drive, etc.) only requires extending that map.
- **Packaging**: \`pyproject.toml\` and \`setup.py\` with the same pattern as GitLab, depending on \`shared-functions\`.
- **Tests**: \`test/test_core.py\` — pointer round-trip and subdata helpers. Auth-mode tests removed along with the static-token methods they covered.
- **Docker**: \`Dockerfile\` and \`connector-githubs\` service in \`docker-compose-build-all.yaml\`.

## Configuration

Required env vars to run the connector:

| Variable | Description |
|---|---|
| \`GITHUB_API_URL\` | GitHub API base URL, e.g. \`https://api.github.com/\` |
| \`GITHUB_SYSTEM_NAME\` | Display name for this source system |
| \`GITHUB_API_VERSION\` | API version header, e.g. \`2022-11-28\` |
| \`GITHUB_CONNECTOR_HOST\` | Bind address |
| \`GITHUB_CONNECTOR_PORT\` | Port |

Optional:

| Variable | Description |
|---|---|
| \`GITHUB_ORG\` | If set, fetches org repos instead of user repos |
| \`GITHUB_EXCLUDE_PATHS\` | Comma-separated path substrings to exclude from indexing |

## What's NOT in scope (for colleagues)

\`handlers.py\` in the search engine needs to receive a \`tokens\` dict from the API layer and pass it to \`connector.get_file_pointers(tokens)\`, \`connector.get_files(tokens)\`, and \`connector.fetch_files(pointers, tokens)\`. Until that's wired up both connectors return empty results for every search.


src/connectors/githubs/.gitignore
```diff
@@ -0,0 +1,8 @@
+__pycache__/
```
jaen22BTH: We have a global .gitignore file at root (https://github.com/Studentprojekt-KCL/document_management_system/blob/develop/.gitignore) make additions here if needed.

src/connectors/githubs/tests/test_core.py
```diff
@@ -0,0 +1,86 @@
...
+                "GITHUB_APP_INSTALLATION_TOKEN": "app-token",
+            },
+            clear=False,
+        ):
+            github = GitHub()
+        self.assertEqual(github.session.headers.get("Authorization"), "Bearer app-token")
+
+
+if __name__ == "__main__":
+    unittest.main()
```
jaen22BTH: Think it might be better to use the pytest entrypoint (should automatically work if this is removed).

src/connectors/githubs/github_collector_api.py
```diff
@@ -0,0 +1,90 @@
...
+    if port is None or not port.isdigit():
+        dms_error(
+            "Port for GitHub connector not set in local environment, please export GITHUB_CONNECTOR_PORT."
+        )
+        return
+    uvicorn.run(api.app, host="0.0.0.0", log_level=api.log_level, port=int(port))
+
+
+if __name__ == "__main__":
+    run()
```
jaen22BTH: Use entrypoint defined in pyproject.toml (already used in Dockerfile).

src/connectors/githubs/github_collector_api.py
```diff
@@ -0,0 +1,86 @@
...
+    if args.dev:
+        api.log_level = "debug"
+
+    port = os.environ.get("GITHUB_CONNECTOR_PORT")
+    if port is None or not port.isdigit():
+        dms_error(
+            "Port for GitHub connector not set in local environment, please export GITHUB_CONNECTOR_PORT."
+        )
+        return
+    uvicorn.run(api.app, host="0.0.0.0", log_level=api.log_level, port=int(port))
```
AlexanderJarnstromBTH: host is hard coded

src/connectors/githubs/interfacer_github.py
```diff
@@ -0,0 +1,427 @@
...
+
+    session: requests.Session
+    api_base: str
+    _binary_skip_logs: int
+    _binary_skip_log_limit: int
+
+    def __init__(self) -> None:
+        """Constructor."""
+        self.session = requests.Session()
+        raw = os.environ.get("GITHUB_API_URL", "https://api.github.com")
```
AlexanderJarnstromBTH: dont think we should have a default value here.

src/connectors/githubs/interfacer_github.py
```diff
@@ -0,0 +1,427 @@
...
+        self.api_base = raw if raw.endswith("/") else raw + "/"
```
AlexanderJarnstromBTH: use `.rstrip()` instead

src/connectors/githubs/tests/test_core.py
```diff
@@ -0,0 +1,82 @@
...
+import base64
+import os
+import unittest
+from unittest.mock import patch
+
+from interfacer_github import GitHub
+from unpacker import unpack_values
+
+
+class TestUnpacker(unittest.TestCase):
```
jaen22BTH: Nice! Although since this is a test for shared_functions methods, should probably be there instead!

src/connectors/githubs/src/interfacer_github.py
```diff
@@ -0,0 +1,430 @@
...
+        size = file.get("size")
+        last_edit = self._last_commit_date_for_path(full_name, path, ref)
+
+        base_structure: dict[Any, Any] = {
+            "metadata": {
+                "unique_pointer": pointer,
+                "name": name,
+                "size": size,
+                "last_edit_date": last_edit,
+                "type": SOURCE_FILE,
```
jaen22BTH, line 216: Add source_system to like:
https://github.com/Studentprojekt-KCL/document_management_system/blob/develop/src/connectors/gitlabs/src/interfacer.py#L160

src/connectors/githubs/src/github_collector_api.py
```diff
@@ -0,0 +1,88 @@
...
+    """Initiate FastAPI using Uvicorn."""
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--dev", action="store_true")
+    args = parser.parse_args()
+
+    api = API()
+    if args.dev:
+        api.log_level = "debug"
+
+    port = os.environ.get("GITHUB_CONNECTOR_PORT")
```
AlexanderJarnstromBTH: preferably use `read_env_variable(env_variable: str)`from [here](https://github.com/Studentprojekt-KCL/document_management_system/blob/develop/src/shared_functions/shared_functions/initialisation_tools.py), for all environment variable grabs except ports which is `read_port(env_variable: str, allowed_ports: tuple[int, ...] = tuple(range(0, 65536)))`

src/connectors/githubs/src/interfacer_github.py
```diff
@@ -0,0 +1,410 @@
...
+        self._binary_skip_logs = 0
+        self._binary_skip_log_limit = 10
+        self._set_default_headers()
+        self._configure_auth()
+
+    def _set_default_headers(self) -> None:
+        self.session.headers.update(
+            {
+                "Accept": "application/vnd.github+json",
+                "X-GitHub-Api-Version": "2022-11-28",
```
AlexanderJarnstromBTH: maybe should be an env variable

src/connectors/githubs/src/interfacer_github.py
```diff
@@ -0,0 +1,410 @@
...
+
+        if not token:
+            dms_info("No GitHub auth token configured; connector runs in public (unauthenticated) mode. (rate limits are lower).")
+            return
+
+        self.session.headers.update({"Authorization": f"Bearer {token}"})
+
+    def _get_repos(self) -> list:
+        """Retrieve all repositories the token can access (user or org)."""
+        org = os.environ.get("GITHUB_ORG")
```
AlexanderJarnstromBTH: do we need to grab it every time the function is called? Constructor maybe?

pyproject.toml
```diff
@@ -17,6 +17,7 @@ select = ["T20", "PYI056", "ASYNC", "B", "SIM", "PL", "UP"]
 disallow_untyped_defs = true
 exclude = '(\.env|venv|__pycache__|\.tox|build/|setup\.py|test/)'
 ignore_missing_imports = true
+explicit_package_bases = true
```
jaen22BTH: Should this check really be skipped? Considering a __init__.py file does exist.

src/connectors/githubs/src/interfacer_github.py
```diff
@@ -0,0 +1,410 @@
...
+    def _incoming_or_legacy_token() -> str | None:
+        incoming = os.environ.get("GITHUB_ACCESS_TOKEN")
+        if incoming:
+            return incoming
+        return os.environ.get("GITHUB_TOKEN")
+
+    @staticmethod
+    def _app_installation_token() -> str | None:
+        # Placeholder for future JWT->installation token exchange.
+        return os.environ.get("GITHUB_APP_INSTALLATION_TOKEN")
```
AlexanderJarnstromBTH: believe we agreed on using (if possible) user sessions given to us from the frontend or users individual access tokens (need to discuss with the rest of the group if needed).

FelonBall: The involved partied should discuss it together. Can we keep the placeholder for now?

src/connectors/githubs/src/interfacer_github.py
FelonBall: Yeeeeah buddy det är rimligt att ha konstruktor på den

pyproject.toml
FelonBall: explicit_package_bases isn't skipping a check; it's needed because mypy runs across multiple packages that each declare package-dir = {"" = "src"}. Without it, mypy derives wrong module names and produces false import errors. The __init__.py doesn't resolve that root ambiguity.

jaen22BTH: Since the packages aren't set as deps in mypy currently, this shouldnt be an issue. However, this will be changed in the future, the solution here should in that case be to create separete tox envs for each service instead. Although, I am not sure what 'root ambiguity is in this context', so I might be wrong. (The only global dep set should be src/shared_functions)

FelonBall: Makes sense. with ignore_missing_imports = true and no cross-package mypy deps, this isn't needed right now. I'll remove it, and we can revisit with per-service tox envs when the time comes.

tox.ini
```diff
@@ -33,11 +33,9 @@ commands = black {toxinidir}/src
 allowlist_externals = pytest
 deps = 
     ./src/connectors/gitlabs
-    ./src/search_engine
```
jaen22BTH, line 36: We would still prefer search engine unittests :D

src/search_engine/README.md
```diff
@@ -85,3 +85,20 @@ Configuration is done through environment variables.
 Optional flags:
 
 - `--dev`, developer mode
+
```
jaen22BTH: This needs to be fully dynamic from the search-engines perspective, please remove this

src/search_engine/src/se_api/services/connector.py
```diff
@@ -41,7 +46,13 @@ def reset(self) -> None:
...
         self.subdata = None
 
-    def get_file_pointers(self) -> list[str]:
+    def _build_auth_headers(self, tokens: dict[str, str] | None) -> dict[str, str]:
+        """Map source-system tokens to their named request headers."""
+        if not tokens:
+            return {}
+        return {self._TOKEN_HEADER_MAP[k]: v for k, v in tokens.items() if k in self._TOKEN_HEADER_MAP and v}
+
+    def get_file_pointers(self, tokens: dict[str, str] | None = None) -> list[str]:
```
jaen22BTH: I believe this function was replaced by 'reindex_needed', but also, am I missing something, or are the tokens never provided to this? (concidering api.py hasnt been updated).

src/search_engine/src/se_api/services/connector.py
```diff
@@ -19,6 +19,11 @@ class Connector:
 
     TIMEOUT: int = 120
 
+    _TOKEN_HEADER_MAP: dict[str, str] = {
+        "github": "X-GitHub-Token",
```
jaen22BTH: We can not have these assumptions hardcoded in the search engine, SE should know nothing about this.

src/connectors/gitlabs/README.md
```diff
@@ -5,6 +5,16 @@ The following needs to be exported in local environment:
     GITLAB_CONNECTOR_BIND_ADDR=<GITLAB_CONNECTOR_BIND_ADDR>
     GITLAB_SYSTEM_NAME=<Name to display in frontend for this Gitlab instance (e.g. 'gitlab').>
 
+## Authentication
```
jaen22BTH: I don't want to be rude, but could we please please please not have LLM generated documentation here, especially considering the description of the code functionality below is incorrect.

src/connectors/gitlabs/src/__init__.py
jaen22BTH, line 1: Why remove this? We do want python to view this as a package.

jaen22BTH (changes requested):
Maybe it would be best not to change other system components in this commit, allowing us to focus on the Github part. Otherwise, large changes in other components, might be better to do this in a separate commit.

tox.ini
FelonBall: Oh, we haven't touched this actually, and the line has never existed on this branch. I'll add it.

src/connectors/gitlabs/src/__init__.py
FelonBall: Both connectors declare package-dir = {"" = "src"} in their pyproject.toml, which already tells setuptools (and the installed package) to treat src/ as the package root. Modules like github_collector_api and interfacer_github are importable directly without an __init__.py. The entry points (github_connector, gitlab_connector) work exactly because of this setup.

The real problem with keeping them: the root pyproject.toml runs mypy across the whole repo. With src/__init__.py in both connector directories, mypy sees two packages both named src and raises "Duplicate module named 'src'".

If we really want to signal that these directories are packages, the clean solution is to rename the inner src/ to a unique name (e.g. github_connector/, gitlab_connector/) and update pyproject.toml accordingly. Happy to do that in a separate PR. For now, the current layout works and passes mypy.

jaen22BTH (approved):

jesperlenn:
LGTJ

PR #294, opened 2026-04-07
Feature/connector system information

jaen22BTH:


src/connectors/gitlabs/src/interfacer.py
```diff
@@ -26,6 +26,7 @@ class GitLabs:
     GIT_BLAME: str = "blame?ref=HEAD"
     GIT_HEAD: str = "?ref=HEAD"
     session: requests.Session
+    SOURCE_SYSTEM: str = "gitlab"
```
AlexanderJarnstromBTH: what if theres two gitlab instances? wouldnt we want to know which and have it as an environment variable? or is that unnecessary?

jaen22BTH: True! Changed to this (implementation will be migrated to shared env var solution in separate commit).

AlexanderJarnstromBTH (approved):
LGTJ

PR #317, opened 2026-04-10
Feature/main api/tokenverification

mamz22:
#331

Adds token-based authentication and authorization to the main API using Keycloak.

All requests to protected endpoints now require a valid bearer access token.

Changes
- Added JWT access token verification using Keycloak JWKS
- Validates Authorization: Bearer <token> header
- Verifies token signature and issuer
- Enforces azp verification 
- Authorizes based on roles: user or admin

Notes
- Audience (aud) is not currently verified (verify_aud=False)
- Look in README for new env var


AlexanderJarnstromBTH (approved):
LGTJ

PR #350, opened 2026-04-14
filLess hardcoded security filters?

EmEhPa:
Only one function to fetch the list of security levels to use in the filters card.


jaen22BTH (approved):
amazing

PR #353, opened 2026-04-14
log-api now sets port according to LOGGER_PORT env var

jesperlenn:


AlexanderJarnstromBTH (approved):
Klart grabben ska bind portar

Issue #347, opened 2026-04-14
Remove hardcoded connected systems and integrate connector endpoint


EmEhPa:
- [x] Displaying for search results
- [x] For filtering

EmEhPa:
Done?

PR #359, opened 2026-04-15
Fetch the source system from metadata #347

EmEhPa:
So it was supposed to only be for resolving the source_system for each file, but decideed to clean up some inn the metadat file.
Only for showing the files, not for the filter card yet.

Update: It works for both now

image


EmEhPa:
Ifal man missade vart filen kommer ifrån

image

EmEhPa:
it works :)

image

jaen22BTH (approved):

PR #380, opened 2026-04-17
This PR adds the ability for admin users to update a document's security classification

muaathaw:
# Add security classification editor for admin users 

- Extract ClassificationEditor.vue as standalone modal component
- Add edit button to SearchPreviewDrawer (admin-only via hasRole)
- Save/cancel flow with toast notifications for feedback
- Auto-reset on document switch or drawer close


src/front_end/src/components/ClassificationEditor.vue
```diff
@@ -0,0 +1,314 @@
...
+
+const props = defineProps({
+  visible: { type: Boolean, default: false },
+  currentLevel: { type: String, default: '' }
+})
+
+const emit = defineEmits(['save', 'cancel'])
+
+/* Security classification levels */
+const SECURITY_LEVELS = ['Public', 'Internal', 'Sensitive', 'Confidential']
```
jaen22BTH: This needs to be dynamic, gathering the classifications from the /classifications. However, this request is already done in SearchFiltersCard.vue, so it is probably unnecessary to make it twice.

src/front_end/src/components/SearchPreviewDrawer.vue
```diff
@@ -24,11 +38,81 @@ const props = defineProps({
...
+const showToast = (success, message) => {
+  if (toastTimer) clearTimeout(toastTimer)
+  toast.value = { visible: true, success, message }
+  toastTimer = setTimeout(() => {
+    toast.value.visible = false
+  }, 4000)
+}
+
+/* Save classification */
+const handleClassificationSave = async (level) => {
```
EmEhPa, line 81: Move to api.js

src/front_end/src/components/SearchPreviewDrawer.vue
```diff
@@ -24,11 +38,81 @@ const props = defineProps({
...
 
 /* Use custom composable to extract metadata for the selected file */
-const { previewTitle, previewType, sourceSystem, previewCreatedAt, previewSize, previewLink, previewSecurityClass } =
+const { previewTitle, previewType, sourceSystem, previewCreatedAt, previewSize, previewLink, previewSecurityClass, uniquePointer } =
   useSearchMetadata(props)
 
 /* AI summary composable */
 const { aiSummaryHtml, summaryError, isGeneratingSummary, generateAISummary } = useAISummary(props)
+
+const API_BASE_URL = window.__ENV__.API_BASE_URL.replace(/\/$/, '')
```
EmEhPa: fetch from api.js

src/front_end/src/components/SearchPreviewDrawer.vue
```diff
@@ -8,9 +8,23 @@
  * <SearchPreviewDrawer :open="isPreviewOpen" :selected-file="selectedFile" :selected-match="selectedMatch" :matches="matches" @close="closePreview" />
  */
 
-import { X, StarsIcon, CalendarDays, HardDrive, FileType2, ExternalLink } from 'lucide-vue-next'
+import { ref, computed, watch } from 'vue'
```
EmEhPa, line 11: Should be possible to only use computed like we do across the system

src/front_end/src/components/SearchPreviewDrawer.vue
```diff
@@ -24,11 +38,81 @@ const props = defineProps({
...
+)
+
+/* Final value used in UI */
+const currentSecurityLevel = computed(() => localSecurityLevel.value)
+
+/* Permissions */
+const canEdit = computed(() => hasRole('admin'))
+
+/* Toast */
+const toast = ref({ visible: false, success: true, message: '' })
```
EmEhPa: What is toast?

src/front_end/src/components/SearchPreviewDrawer.vue
```diff
@@ -24,11 +38,81 @@ const props = defineProps({
...
+    isEditingClassification.value = false
+  } catch (err) {
+    showToast(false, `Update failed: ${err.message}`)
+  } finally {
+    classificationEditorRef.value?.resetSaving()
+  }
+}
+
+/* Reset UI on change */
+watch(
```
EmEhPa, line 112: Computed?

EmEhPa (changes requested):
Cant go through everything now but these are some comments

src/front_end/src/components/SearchPreviewDrawer.vue
muaathaw: This is a side effect — it resets the editor and notification when the user switches documents. Computed only works for derived values, not for triggering actions.

src/front_end/src/components/SearchPreviewDrawer.vue
muaathaw: Status notification — briefly shown after save/cancel to confirm the result

src/front_end/src/components/SearchPreviewDrawer.vue
muaathaw: unfortunately in this case it´s not possible to use only computed, each does a different job: computed for derived values like canEdit and currentSecurityLevel, ref for interactive state the user changes (modal open/close, notification visibility), and watch to reset that state as a side effect when the document changes. 
computed can't trigger side effects or hold mutable state.

muaathaw:
the code is almost done need to test enough and Oscar wanted to change a little in session tokens.
If you see it Don´t merge now

src/front_end/src/components/SearchPreviewDrawer.vue
kebabohvo: fixed

src/front_end/src/components/SearchPreviewDrawer.vue
kebabohvo: has been renamed to notification

src/front_end/src/components/SearchPreviewDrawer.vue
kebabohvo: fixed

kebabohvo:
This pr will be merged once a small change is made, we already have a security class. The changes needed for the pr to be accepted is the following:
1. move the button to beside the Security Class in Technical METADATA
2. remove the security classification from above Technical Metadata.
Screenshot from 2026-04-22 17-14-01

EmEhPa:
Nice, the logic is correct and it does its job. However, currently it shows the classification 3 time in total, which is a bit overkill. 

image

Untitled(1)

So what you could do is 
changes

EmEhPa:
So it seems lite it does change the security level but the user need to sign out and sign in again in order for the level to change in the display. Is it possible to make so it change directly?

(Could the cell for classification be the coulor of the level?)

EmEhPa (approved):

PR #389, opened 2026-04-20
DMIS API: replaced requests to httpx

mamz22:
Migrate dmis api from requests to httpx AsyncClient


src/api/src/dmis_api/api.py
```diff
@@ -118,22 +127,21 @@ async def execute_post_request(self, url: str, request: Request) -> JSONResponse
...
+            response = await self.http_client.post(
                 url,
                 params=params,
                 json=body,
-                timeout=120,
             )
             response.raise_for_status()
             response_data = response.json()
-        except requests.JSONDecodeError as exc:
+        except ValueError as exc:
```
jaen22BTH: Does the json method in httpx not raise the jsondecodeerror from json standard lib?

mamz22: It raises jsondecodeerror which is a subclass of valueerror. I can change it to jsondecodeerror if u would like it to be more explicit?

jaen22BTH: Not to important, however, we do not catch parents in other sections. What was the reason to migrate to it here?

mamz22: The reason was that jsondecodeerror enherits from valueerror, so it works either way

jaen22BTH: Then why not catch Exception everywhere?

jaen22BTH (approved):

PR #393, opened 2026-04-20
Feature/ranker rework

FilipDar:
Complete ranker rework with new logic

vector database support.


AlexanderJarnstromBTH (approved):
fix error later LGTE

PR #401, opened 2026-04-20
fix/github-optimization: configurable timeouts, shared HTTP client, and streaming fixes for GitHub

FelonBall:
## Summary

- `NUM_WORKERS`, and `SHARED_CLIENT` are now read
  from required env vars instead of being hardcoded constants
- Adds `_http_client()` context manager: when `SHARED_CLIENT=true`
  a single `httpx.AsyncClient` is shared across all download workers (one connection
  pool); when false, each worker owns its own client
- Switches zip download URL from semi-hardcoded url to the
  `/zipball/` API endpoint via `api_base` — works automatically for GitHub Enterprise
- Moves `_get_repos` off the event loop via
  `asyncio.to_thread` to avoid blocking during the initial repo fetch
- Fixes `task_done()` not being called for the sentinel item in `_download_files`
  on both connectors — previously `queue.join()` would hang
- Marks `/files_to_index` as deprecated (`deprecated=True` + `warnings.warn`);
  endpoint is kept until the connector gateway is in place
- Extends test suites: 8 new GitHub tests


src/connectors/github/src/interfacer_github.py
```diff
@@ -5,37 +5,41 @@
...
-import requests
+import httpx
 
 from shared_functions.variables import SOURCE_FILE
 from shared_functions.unpacker import unpack_values
 from shared_functions.dmis_logger import dms_error, dms_info, dms_warning
 from shared_functions.initialisation_tools import read_env_variable
 
 HTTP_OK = 200
+REQUEST_TIMEOUT = 120
```
Sjomullen, line 29: env vars maybe?

src/connectors/github/src/interfacer_github.py
```diff
@@ -5,37 +5,41 @@
...
 
 
 class GitHub:
     """GitHub connector."""
 
-    session: requests.Session
+    NUM_WORKERS: int = 10
```
Sjomullen: env vars?

src/connectors/github/src/interfacer_github.py
```diff
@@ -262,12 +249,105 @@ def _unpack_zip(self, content: bytes, full_name: str, branch: str) -> list:
...
+                continue
+            ts_obj = datetime.fromisoformat(ts.replace("Z", "+00:00"))
+            if ts_obj <= provided_date:
+                continue
+            latest_update = max(latest_update, ts_obj)
+            branch = repo.get("default_branch")
+            if not isinstance(branch, str):
+                branch = self._default_branch_for_repo(fn, token)
+            owner, _, name = fn.partition("/")
+            zip_url = f"https://codeload.github.com/{owner}/{name}/zip/refs/heads/{branch}"
```
Sjomullen: should this url really be hardcoded and not env vars?

FelonBall:
LGTJ?

jaen22BTH (changes requested):
Please revert all changes made to the GitLab instance, I believe spawning a new thread in the unzip method in unnecessary, but if false, this should be done in a separate commit.

src/connectors/github/README.md
```diff
@@ -5,6 +5,9 @@ The following must be exported in the local environment:
     CONGITHUB_GITHUB_API_URL=<GITHUB_API_BASE_URL>        # e.g. https://api.github.com/
     CONGITHUB_GITHUB_SYSTEM_NAME=<SOURCE_SYSTEM_NAME>
     CONGITHUB_GITHUB_API_VERSION=<API_VERSION>            # e.g. 2022-11-28
+    CONGITHUB_REQUEST_TIMEOUT=<SECONDS>                   # e.g. 120
```
jaen22BTH: This was removed?

src/connectors/github/README.md
```diff
@@ -5,6 +5,9 @@ The following must be exported in the local environment:
...
+    CONGITHUB_NUM_WORKERS=<COUNT>                         # e.g. 10
```
jaen22BTH: Should this rly be an env variable?

FelonBall: Maybe not. I could change it to a constant instead. What is a reasonable number? 10?

src/connectors/github/src/interfacer_github.py
```diff
@@ -5,54 +5,57 @@
...
 HTTP_OK = 200
+REQUEST_TIMEOUT = 120
 
 
-class GitHub:
+class GitHub:  # pylint: disable=too-many-instance-attributes
     """GitHub connector."""
 
-    session: requests.Session
+    client: httpx.Client
```
jaen22BTH: Doing this raises red flags, for me atleast, that future implementations like:
client.headers[] is set,
or the server sets cookies in this clients jar, meaning they might bleed into next session.

What is the thinking here?
if we want this structure, I guess we could always use like a/the request wrapper, and name this '_client' to make it clear. And I belive we can do something like
_client = httpx.Client(cookies=None)

src/connectors/github/src/interfacer_github.py
```diff
@@ -262,12 +246,123 @@ def _unpack_zip(self, content: bytes, full_name: str, branch: str) -> list:
...
+
+    async def _download_files(
+        self, task_queue: asyncio.Queue, zip_queue: asyncio.Queue, client: httpx.AsyncClient | None, token: str | None
+    ) -> None:
+        """Download repo archives from task_queue and put raw bytes into zip_queue."""
+        headers = {"Authorization": f"Bearer {token}"} if token else {}
+        owned = client is None
+        if owned:
+            client = httpx.AsyncClient(timeout=REQUEST_TIMEOUT, follow_redirects=True)
+        assert client is not None
```
AlexanderJarnstromBTH: `if client is (not) None:`

src/connectors/github/src/interfacer_github.py
```diff
@@ -340,15 +410,24 @@ def _provided_date(subdata: str | None) -> datetime:
             dms_error("GitHub connector could not interpret subdata: %s", subdata)
             return datetime.min.replace(tzinfo=timezone.utc)
 
+    def _request(self, url: str, token: str | None = None) -> httpx.Response:
+        """Build a clean, stateless GET request — all headers explicit, no cookie carry-over."""
```
jaen22BTH: Is it ok if we remove the —, and maybe we should say like 'Execute a clean[...]'?

src/connectors/github/src/interfacer_github.py
```diff
@@ -262,12 +246,123 @@ def _unpack_zip(self, content: bytes, full_name: str, branch: str) -> list:
...
+        """Streaming equivalent of files_to_index: yields subdata header then one file per chunk."""
+        task_queue: asyncio.Queue = asyncio.Queue()
+        zip_queue: asyncio.Queue = asyncio.Queue()
+        output_queue: asyncio.Queue = asyncio.Queue()
+
+        repos = await asyncio.to_thread(self._get_repos, token)
+        latest_update = await self._enqueue_repos(repos, self._provided_date(subdata), token, task_queue)
+        generated_subdata = base64.urlsafe_b64encode(latest_update.isoformat().encode()).decode()
+
+        async with self._http_client() as client:  # pylint: disable=contextmanager-generator-missing-cleanup
```
AlexanderJarnstromBTH: ignoring pylint D:

src/connectors/github/src/interfacer_github.py
```diff
@@ -5,55 +5,53 @@
...
 from shared_functions.initialisation_tools import read_env_variable
 
 HTTP_OK = 200
+REQUEST_TIMEOUT = 120
+NUM_WORKERS = 10
+SHARED_CLIENT = True
 
 
-class GitHub:
+class GitHub:  # pylint: disable=too-many-instance-attributes
```
AlexanderJarnstromBTH: woah again

src/connectors/github/src/interfacer_github.py
```diff
@@ -262,12 +246,123 @@ def _unpack_zip(self, content: bytes, full_name: str, branch: str) -> list:
...
         return files_data
 
+    @asynccontextmanager
+    async def _http_client(self) -> AsyncGenerator[httpx.AsyncClient | None, None]:
+        """Yield a shared AsyncClient when SHARED_CLIENT=True, else yield None per worker."""
+        if SHARED_CLIENT:
+            async with httpx.AsyncClient(timeout=REQUEST_TIMEOUT, follow_redirects=True) as client:
+                yield client
+        else:
+            yield None
```
AlexanderJarnstromBTH: Why have a async generator yield a AsyncClient? into a with?? inside a with??? dubble with????

jaen22BTH (approved):

PR #407, opened 2026-04-21
Feature/rerank mk67

FilipDar:
Ranker support to deliver metadata

### DIFF
--- a/src/stochastic_analyzer/gateway/services/connector.py
+++ b/src/stochastic_analyzer/gateway/services/connector.py
@@ -77,10 +77,6 @@ class Connector:
     async def get_file_metadata(self, pointers: list[str]) -> list[dict]:
         """Fetch file metadata from the connector without content payloads.
 
-        The connector applies authorization filtering: files the requesting
-        user is not permitted to view are omitted from the response, so the
-        returned list may be shorter than the input pointers list.
-
         Args:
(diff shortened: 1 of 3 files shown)


src/stochastic_analyzer/gateway/services/connector.py
```diff
@@ -6,7 +6,6 @@
 import httpx
 
 from gateway.schemas import InputItem, MetadataTemplate
-
```
AlexanderJarnstromBTH, line 9: where it at

src/stochastic_analyzer/gateway/services/connector.py
```diff
@@ -74,3 +73,32 @@ async def get_file_contents(self, pointers: list[str]) -> list[InputItem]:
...
+            pointers: List of unique file pointers.
+
+        Returns:
+            List of metadata dicts as provided by the connector, or an
+            empty list if the request fails.
+        """
+        async with httpx.AsyncClient() as client:
+            try:
+                response = await client.post(
+                    f"{self.url.rstrip('/')}/get_files",
```
AlexanderJarnstromBTH, line 90: why strip every call when can do it on get?

AlexanderJarnstromBTH (approved):
LGTJ

PR #412, opened 2026-04-22
Fix/api pass auth token

jaen22BTH:


EmEhPa (approved):
Should work, we need to remember to remove that if statement

PR #413, opened 2026-04-22
Frontend/wished fixes

EmEhPa:
- Hamburger icon is aligned with others in sidebar
- Logo redirects to search
- Searchbar icon is the same size as the other icons
- Not two line if choosing security & compilance

Wishes from Filip and Vlan


kebabohvo (approved):
Looks ok,

PR #414, opened 2026-04-22
Feature/reload page fix

kebabohvo:
Changes:
made states persistent.
made useReload.js that sync's local Storage to initial state.
clears state upon logout.


EmEhPa (approved):
Det r bra, du vet vad du ska göra härnäst :)

PR #415, opened 2026-04-22
Gitlab auth service

jaen22BTH:
Gitlab auth_user and callback endpoints for catching session GitLab tokens. Refresh endpoint for refreshing token. New shared functions for HMAC signing payloadds with priv key.


AlexanderJarnstromBTH (approved):
LGTJ

PR #425, opened 2026-04-23
Feature/samba connector/base

AlexanderJarnstromBTH:


src/connectors/samba/src/services/samba.py
```diff
@@ -0,0 +1,322 @@
...
+        Mounts the share as the requesting user and grabs the requested files.
+
+        Args:
+            content: json body.
+            include_content: to include file content or not.
+            include_last_edit_date: to include the modification date.
+        Returns: list of files.
+        """
+        pointers: list[str] | None = content.get("file_pointers")
+        username: str | None = content.get("username", self.mount_options.user)  # DO NOT KEEP
```
jaen22BTH: Do not keep as in now, or this should be removed later? If so, maybe we should begin to have some good keyword in the comments for those implementations

tox.ini
```diff
@@ -24,7 +25,7 @@ commands = ruff check {toxinidir}/src
 [testenv:mypy-api]
 setenv =
     MYPYPATH = {toxinidir}/src/shared_functions/logger
-commands = mypy src
+commands = mypy --explicit-package-bases src
```
jaen22BTH: Is this needed since all packages should have an \__init__.py file?

If so, place in pyproject.toml:
[mypy]
explicit_package_bases = True

jaen22BTH (approved):

jaen22BTH (commented):
Update env var names

jaen22BTH (approved):

PR #434, opened 2026-04-27
Feature/connector gateway

jesperlenn:


jaen22BTH (approved):
Looks really good! Do we want to migrate away from httpx here? Or will this be done at a later stage if decided?

Issue #367, opened 2026-04-16
Documents only toggle


jaen22BTH:
It was decided that the term for excluding 'code files', and only excluding 'text files' will be 'Documents only'

What exactly this implies will need to be implemented, together with a parameter and delivery mechanism.

EmEhPa:
Done

PR #444, opened 2026-04-28
Feature/frontend/toggle

EmEhPa:
Refers to: #367

Document Only state: 
image

The dropdown shows all file types defined, while the filtercard only shows the 11 most common. 
image

If the user press the button after one search, the request will be sent down again with changed boolean value, nice :)


jaen22BTH (approved):
Amazing eller?

AlexanderJarnstromBTH:
amazing eller?

PR #457, opened 2026-04-29
Feature/http only cookies

kebabohvo:
Changes Made:

#FRONTEND RELATED FILES

#SearchView.vue:
- reverted back to non persistant state

#LoginView.vue:
- placed isLoading within to indicate that we are currently signing in, disabling the button.

#ErrorStatusView.vue:
- Removed isLoggedIn, we no longer handle tokens or determine login state locally, it is handled by mainAPI.
- Simpler check if to go to login, if not authenticated you should go to login.

#AuthCallBack.vue:
- Removed unnecessary local Storage keys, we have switched to cookies, and tokens are handled by mainAPi.
- We keep pkce_verifier and oidc state temporarily, due to their need for connecting with Keycloak and getting the exchange code.
- Removed Tokens from frontend, moved to using cookies, which are set and handled by backend (mainAPI).
- Calls isAuthenticated(), an endpoint for checking authentication before moving to search, to make sure cookies are set properly.

#config.js:
- Removed keycloakTokenUrl since we don't use it
- Removed session_keys (ACCESS_TOKEN, ID_TOKEN, REFRESH_TOKEN) Since we don't handle tokens any longer.
- We kept OIDC and PKCE due to OAuth thingy with keycloak.

#AuthClient.js:
- A file dedicated for cookie-based authentication calls.
- Added exchangeAuthorizationCode which sends exchange code and pkce verifier to mainAPI for it to then exchange them for tokens.
- Added isAuthenticated which provides a simple check if the user is authenticated, used by router guards and login validation, uses endpoint /auth/check.
- Added getCurrentUser which returns users info (roles, etc) uses endpoint /auth/me.

#auth.js:
- Removed all JWT parsing, now managed by mainAPI through http-only cookies.
- Removed helpers that aren't used anylonger, (isLoggedIn, getAccessToken, getRefreshToken, saveTokens, isTokenExpired, refreshToken)
- Replaced front-end role checks with a call to getCurrentUser which returns info about the current user.
- hasRole now uses authClient's getCurrentUser to get info about the user.
- Refresh of tokens is now moved to backend, now only calls for and checks if refresh was successful.
- logout is now updated and call logout endpoint in mainAPI that handles logout through keyclaok.

#api.js:
- Removed SESSION_KEY_ACCESS_TOKEN import and bearer token injection.
	- Frontend no longer sends tokens, we send cookies with each http communication.
- replaced authFetch which was token-based with apiFetch which is cookie-based.
- Added a few paths:
	- authCheck
	- authMe
	- authRefresh
	- authLogout
	- codeExchange
- Added AuthFetch as an alias for apiFetch due to some using the old code.

#index.js:
- Replaced token-based route guarding with backend cookie based authentication.
- removed Session_keys.
- our route guard now is dependent on server calls, which is why we use async on before each now.

#useAuthSession.js:
- Removed all old token based logic.
- Added broadcastActivity on mount to fix users not being intialized timer upon opening app.
- Removed token inspections etc.

#aiSummary.js:
- Reverted back to non persistent state

#aiRerank.js:
simliar change and as in aiSummary.js.

#TheSideBar.vue:
- replaced hasRole with getCurrentUser + computed role check.
- stores results from getCurrentUser from  /auth/me endpoint.
- Added Watch on route.fullPath to reload user roles dynamically.

#TheHeader.vue:
- modified the file to use our current logic
- Added a watcher to look at change in roles

#App.vue:
- Added '/auth/callback' route to isPublicErrorPage
	- fixes layout issues during OAuth flow

#API RELATED FILES

#authRoutes.py:
- Added AuthRoutes to centralize authentication endpoints in mainAPI.
- moved Oauth code exchange to backend.
- Stores access_token, refresh_token, id_token as http only cookies.
- added support for /auth/me, /auth/check, /auth/refresh, /auth/logout.
- Added helper methods for token request, cookie setitng, cookie-based token verification.
- removed need for frontend token storage and decoding.

#api.py:
- Added necessary additions for authRoutes to work.


src/api/src/dmis_api/auth.py
```diff
@@ -15,7 +15,7 @@
 class TokenVerifier:
     """Verify OAuth2/OIDC bearer access tokens and enforce audience, azp and scope-based authorization."""
 
-    def __init__(
+    def __init__(  # pylint: disable=too-many-arguments, too-many-positional-arguments, too-many-locals
```
mamz22: This wasn't needed before, why added ?

src/api/src/dmis_api/api.py
```diff
@@ -45,11 +53,31 @@ def __init__(
...
+            keycloak_logout_url=keycloak_logout_url,
+            frontend_client_id=frontend_client_id,
+            frontend_redirect_uri=frontend_redirect_uri,
+        )
 
         self.app.add_exception_handler(
             RequestValidationError,
             self.validation_exception_handler,
         )
+        self.app.add_middleware(
```
mamz22: If not mistaken this is only used for local development, so is it really necessary to leave this here to get pushed to the development branch? Could be wrong

src/api/src/dmis_api/auth_routes.py
```diff
@@ -0,0 +1,208 @@
...
+from fastapi import Cookie, Form
+from fastapi.responses import JSONResponse, RedirectResponse
+
+# pylint: disable=missing-function-docstring,broad-exception-caught
+
+
+class AuthRoutes:
+    """Authentication route handlers."""
+
+    ACCESS_COOKIE_MAX_AGE = 3600
```
mamz22, line 20: Would put this as a env var

src/api/src/dmis_api/api.py
```diff
@@ -10,15 +10,19 @@
...
+from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse
 
 from shared_functions.dmis_logger import dms_warning, dms_info
 from shared_functions.initialisation_tools import read_env_variable, read_port
 
 from .auth import TokenVerifier
+from .auth_routes import AuthRoutes
+
+# pylint: disable=too-many-arguments,too-many-positional-arguments,too-many-locals,missing-function-docstring,pointless-string-statement
```
mamz22: Probably can go around some of the pylint varnings in other ways than using this line

src/api/src/dmis_api/auth_routes.py
```diff
@@ -0,0 +1,208 @@
...
+        key: str,
+        value: str,
+        max_age: int,
+    ) -> None:
+        response.set_cookie(
+            key=key,
+            value=value,
+            httponly=True,
+            secure=True,
+            samesite="none",
```
mamz22, line 64: would maybe consider to put this as "lax" or "strict"

src/api/src/dmis_api/api.py
kebabohvo: yeah removing it.

src/api/src/dmis_api/api.py
```diff
@@ -237,6 +313,10 @@ def run() -> None:
         "stochan": read_env_variable("DMISAPI_STOCHAN_URL"),
         "congateway": read_env_variable("DMISAPI_CONGATEWAY_URL"),
     }
+    frontend_client_id = read_env_variable("FRONTEND_AD_CLIENT_ID")
```
jaen22BTH: Change env var names :D

src/api/src/dmis_api/api.py
```diff
@@ -68,6 +96,12 @@ def __init__(
         self.app.add_api_route("/connector/{endpoint}", self.connector_get, methods=["GET"])
         self.app.add_api_route("/connector/{endpoint}", self.connector_post, methods=["POST"])
 
+        self.app.add_api_route("/auth/codeExchange", self.auth_routes.code_exchange, methods=["POST"])
+        self.app.add_api_route("/auth/check", self.auth_routes.check_auth, methods=["GET"])
+        self.app.add_api_route("/auth/me", self.auth_routes.auth_me, methods=["GET"])
+        self.app.add_api_route("/auth/refresh", self.auth_routes.refresh_auth, methods=["POST"])
+        self.app.add_api_route("/auth/logout", self.auth_routes.logout_auth, methods=["GET"])
```
mamz22: could be wrong, but shouldn't this be a POST request to prevent CSRF

src/api/src/dmis_api/auth_routes.py
```diff
@@ -0,0 +1,208 @@
...
+    async def _request_tokens(self, data: dict[str, str]) -> dict[str, Any] | None:
+        try:
+            resp = await self.http_client.post(
+                self.keycloak_token_url,
+                data=data,
+                headers={"Content-Type": "application/x-www-form-urlencoded"},
+            )
+            resp.raise_for_status()
+            return resp.json()
+        except Exception:
```
mamz22: A very broad exception consider to have a more specific exception

src/api/src/dmis_api/auth_routes.py
```diff
@@ -0,0 +1,208 @@
...
+        self.keycloak_logout_url = keycloak_logout_url
+        self.frontend_client_id = frontend_client_id
+        self.frontend_redirect_uri = frontend_redirect_uri
+
+    def _verify_cookie_token(self, access_token: str | None) -> dict[str, Any] | None:
+        if not access_token:
+            return None
+        try:
+            return self.token_verifier.verify_access_token(f"Bearer {access_token}")
+        except Exception:
```
mamz22: A very broad exception consider to have a more specific exception

src/api/src/dmis_api/auth_routes.py
```diff
@@ -0,0 +1,208 @@
...
+
+    async def _request_tokens(self, data: dict[str, str]) -> dict[str, Any] | None:
+        try:
+            resp = await self.http_client.post(
+                self.keycloak_token_url,
+                data=data,
+                headers={"Content-Type": "application/x-www-form-urlencoded"},
+            )
+            resp.raise_for_status()
+            return resp.json()
```
mamz22: maybe should consider to validate that the response is a dict and can handle invalid JSON with for e.g. isinstance()

src/api/src/dmis_api/auth_routes.py
```diff
@@ -0,0 +1,208 @@
...
+from fastapi.responses import JSONResponse, RedirectResponse
+
+# pylint: disable=missing-function-docstring,broad-exception-caught
+
+
+class AuthRoutes:
+    """Authentication route handlers."""
+
+    ACCESS_COOKIE_MAX_AGE = 3600
+    REFRESH_COOKIE_MAX_AGE = 30 * 24 * 3600
```
mamz22, line 21: Would put this as a env var

src/api/src/dmis_api/auth_routes.py
kebabohvo: yeah, forgot to change that, had lax previously.

src/api/src/dmis_api/api.py
kebabohvo: yes boss ! 
0_0

src/api/src/dmis_api/auth.py
kebabohvo: will remove it, came due to lint checking where they were, forgot to remove. Sorry.

src/api/src/dmis_api/auth_routes.py
kebabohvo: True

src/api/src/dmis_api/auth_routes.py
kebabohvo: True

src/api/src/dmis_api/api.py
kebabohvo: Yeah, forgot to remove it, was using to test locally.

src/api/src/dmis_api/api.py
kebabohvo: fixed.

src/api/src/dmis_api/api.py
```diff
@@ -44,6 +50,14 @@ def __init__(
...
         self.token_verifier = token_verifier
+        self.auth_routes = AuthRoutes(
+            token_verifier=self.token_verifier,
+            http_client=self.http_client,
+            keycloak_token_url=keycloak_token_url,
+            keycloak_logout_url=keycloak_logout_url,
+            dmisapi_client_id=dmisapi_client_id,
+            dmisapi_redirect_uri=dmisapi_redirect_uri,
+        )
         self.http_client = None
```
mamz22, line 48: move this before self.auth_routes please, its being used before its set in a instance

src/api/src/dmis_api/auth_routes.py
```diff
@@ -0,0 +1,218 @@
+"""Authentication routes for DMIS API."""
+
+from __future__ import annotations
+
+from typing import Any
+from urllib.parse import urlencode
+
+import httpx
```
mamz22: shouldnt be used

mamz22: please migrate everything related to httpx to aiohttp

src/api/src/dmis_api/auth_routes.py
mamz22: still needs to be changed

src/api/src/dmis_api/auth_routes.py
mamz22: still needs to be changed

src/api/src/dmis_api/auth_routes.py
mamz22: still needs to be changed

src/api/src/dmis_api/api.py
```diff
@@ -32,10 +33,15 @@ class API:
     required_scopes: dict[str, list[str]]
     http_client: aiohttp.ClientSession | None
 
+    # pylint: disable=too-many-arguments,too-many-positional-arguments
```
mamz22: is it possible to resolve this in a different way?

src/api/src/dmis_api/api.py
```diff
@@ -236,6 +289,7 @@ async def connector_post(
         return await self.execute_post_request(f"{self.upstream_urls['congateway']}/{endpoint}", request, authorization)
 
 
+# pylint: disable=too-many-locals
```
mamz22: Could we change the code instead of using this line?

src/api/pyproject.toml
```diff
@@ -12,4 +12,7 @@ dynamic = ["dependencies"]
 package-dir = {"" = "src"}
 
 [project.scripts]
-dmis_api = "dmis_api.api:run"
\ No newline at end of file
+dmis_api = "dmis_api.api:run"
+
+[tool.ruff.lint]
```
jaen22BTH: Pls remove this

src/api/src/dmis_api/auth_routes.py
kebabohvo: fixed on next push

src/api/src/dmis_api/auth_routes.py
kebabohvo: we will fix this when we fix the use of refresh tokens in a new branch.

src/api/src/dmis_api/api.py
kebabohvo: fixed in next push

src/api/src/dmis_api/auth_routes.py
kebabohvo: fixed in next push

src/api/pyproject.toml
kebabohvo: fixed in next push

jaen22BTH (approved):

PR #458, opened 2026-04-29
Auth user endpoint callback recieves callback addr in header

jaen22BTH:


jesperlenn (approved):
Underskatta inte sladdhajen

PR #461, opened 2026-04-29
Reverted subdata to project based structure and made index stream end…

jaen22BTH:
…point streaming


AlexanderJarnstromBTH (approved):
lgtj

PR #462, opened 2026-04-30
Feature/query/unit tests

01alireza:
Things that came with this commit:

**test_classifier.py -** 
**test_connector.py -**
**test_routes.py -**
**test_schemas.py -** 
**test_summarizer.py -** 
**test_md_pdf.py -** 
**setup for the unit test -**


src/stochastic_analyzer/test/test_classifier.py
```diff
@@ -0,0 +1,357 @@
...
+        """Constructor stores the injected client."""
+        mock_client = AsyncMock(spec=httpx.AsyncClient)
+        c = Classifier(url="http://tei", escalation_threshold=0.05, client=mock_client)
+        self.assertIs(c.client, mock_client)
+
+    def test_class_level_defaults(self) -> None:
+        """Class-level constants have expected default values."""
+        self.assertEqual(Classifier.max_chars, 2000)
+        self.assertEqual(Classifier.batch_size, 32)
+        self.assertEqual(Classifier.timeout, 15.0)
```
Sjomullen: Why? Seems pretty useless to check for values that the customer or we might changes.. Would be better to just have a check if they have a value if even check for it at all

src/stochastic_analyzer/test/test_classifier.py
```diff
@@ -0,0 +1,357 @@
...
+        """Premise includes the author from metadata."""
+        item = _make_item(author="Bob")
+        result = self.classifier._build_inputs([item])
+        self.assertIn("Bob", result[0][0])
+
+    def test_premise_contains_content(self) -> None:
+        """Premise includes the document content."""
+        item = _make_item(content="Top secret data.")
+        result = self.classifier._build_inputs([item])
+        self.assertIn("Top secret data.", result[0][0])
```
Sjomullen: Why do we test if there is author, document name and content? That is out of our scope for our part of this project, we cannot control if the creators of the document includes everything and some files might be corrupt or just missing content (unlikely..)

src/stochastic_analyzer/test/test_classifier.py
```diff
@@ -0,0 +1,357 @@
...
+
+    async def test_post_called_with_correct_url(self) -> None:
+        """classify() posts to <url>/predict."""
+        classifier = _make_classifier(url="http://tei-host", threshold=0.0)
+        _mock_post_response(classifier.client, [0.1, 0.2, 0.3, 0.9])
+
+        await classifier.classify([_make_item()])
+
+        called_url = classifier.client.post.call_args[0][0]
+        self.assertEqual(called_url, "http://tei-host/predict")
```
Sjomullen: We use /classify and not /predict

src/stochastic_analyzer/test/test_summarizer.py
```diff
@@ -0,0 +1,347 @@
...
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _make_lang_config(sample_size: int = 100, threshold: int = 2) -> LanguageConfig:
+    return LanguageConfig(sample_size=sample_size, swedish_char_threshold=threshold)
+
+
+def _make_config(
+    url: str = "http://llm",
```
Sjomullen: isn't this supposed to be the url for the gpu srv?

src/stochastic_analyzer/test/test_summarizer.py
```diff
@@ -0,0 +1,347 @@
...
+    model: str = "ministral",
```
Sjomullen: this should be ministral-3:14B

src/stochastic_analyzer/test/test_summarizer.py
```diff
@@ -0,0 +1,347 @@
...
+    timeout: int = 30,
```
Sjomullen: What is this timeout?

src/stochastic_analyzer/test/test_summarizer_pdf.py
```diff
@@ -0,0 +1,157 @@
...
+        mock_pdf_cls.return_value = mock_pdf
+        mock_pdf.save_bytes.side_effect = lambda buf: buf.write(b"%PDF-fake")
+
+        md = "# My Section\nBody text."
+        self.converter.convert(md)
+
+        mock_section_cls.assert_called_once_with(md)
+
+
+if __name__ == "__main__":
```
jaen22BTH, line 191: Pls remove this.

tox.ini
```diff
@@ -36,12 +36,14 @@ deps =
     ./src/connectors/gitlab
     ./src/connectors/github
     ./src/search_engine
+    ./src/stochastic_analyzer
     pytest>=9.0
+    httpx
```
jaen22BTH: External dependencies should not be installed here, should be part of package.

src/stochastic_analyzer/test/test_classifier.py
01alireza: These tests aren't checking whether the incoming document has an author or name. They're verifying that _build_inputs correctly includes those fields in the NLI premise string when they are present. That's entirely within our scope since _build_inputs is our code. If someone accidentally refactors that format string and drops the author field, classification quality would silently degrade with no test catching it. The missing-data cases (corrupt/missing fields) are already covered separately by test_missing_name_falls_back_to_unknown and test_missing_author_falls_back_to_unknown, so both sides are tested.

classifier.py line 59 - 70.

    def _build_inputs(self, items: list[InputItem]) -> list[list[str]]:
        """Build NLI premise-hypothesis pairs for all documents."""
        inputs = []
        for item in items:
            **doc_name = item.metadata.name or "Unknown Document"
            author = item.metadata.author or "Unknown Author"
            rich_context = f"Name: {doc_name}. Author: {author}. Content: {item.content[:self.max_chars]}"**

            for trigger in LABEL_TRIGGERS:
                inputs.append([rich_context, trigger])

        return inputs

src/stochastic_analyzer/test/test_classifier.py
01alireza: The classifier currently posts to /predict (see classifier.py, fetch_batch). The test matches the implementation. If we're changing the endpoint to /classify, that change needs to happen in the source first, then the test would follow.

Classifier.py line 105 -111.

        async def fetch_batch(start_idx: int, batch: list[list[str]]) -> None:
            response = await self.client.post(
                f"{self.url}/predict",
                json={"inputs": batch},
                timeout=self.timeout,
            )
            response.raise_for_status()

src/stochastic_analyzer/test/test_summarizer.py
01alireza: We use timeout in the summarizer, and the test inherits the same value:

Summerizer line 51 - 57:

class SummarizerConfig:
    """Configuration for the Summarizer service."""

    url: str
    model: str
    timeout: int  # HTTP request timeout in seconds
    lang_config: LanguageConfig

PR #464, opened 2026-04-30
Bug/gateway connector/response

AlexanderJarnstromBTH:
- Hot fix for error: `Invalid args for response field! Hint: check that starlette.responses.RedirectResponse | None is a valid Pydantic field type.`
- `read_env_variable` (`required=True` is only there in case default behavior changes.) should only return str.


jesperlenn (approved):
LGTM from my phone

Issue #493, opened 2026-05-05
Display unique pointer


jaen22BTH:
For services such as SMB where no clickable URL exists, there is no way of knowing which precise file it is referenced. In this case, the unique pointer should probably be displayed in the frontend.

EmEhPa:
Should do aother thing here aswell

EmEhPa:
Would say this is donw now

PR #506, opened 2026-05-05
Feature/frontend/smb clickable url 

EmEhPa:
Refers to: #493

Ehh, so for SMB there is no clickable URL link so instead we have the sulotion to copy the unique pointer. The flow works. 

During waiting time, I decideed to update the README for the structure. Not much but it was a bit outdates :)


jaen22BTH (approved):
LATJ eller?

PR #519, opened 2026-05-06
fixed roles access based control. A bug previously allowed users to l…

kebabohvo:
small pr


src/api/src/dmis_api/auth_routes.py
```diff
@@ -104,6 +104,13 @@ async def check_auth(self, access_token: str | None = Cookie(default=None)) -> J
         if not claims:
             raise HTTPException(status_code=401)
 
+        client_roles = self._get_client_roles(claims)
+        if not client_roles:
+            raise HTTPException(
+                status_code=403,
+                detail="User does not belong to dms-frontend-dev",
```
mamz22: I believe we shouldn't hardcode the azp name because the stakeholders can change their azp name which could cause confusion. Also maybe shouldn't specify the azp name in the detail code unless there is a reason for it of course

src/front_end/src/components/SearchPreviewDrawer.vue
```diff
@@ -57,6 +57,12 @@ const { aiSummaryHtml, summaryError, isGeneratingSummary, generateAISummary } =
 /* Rerank */
 const { aiRerankResultsComputed, isReranking, rerankError, generateAIRerank } = useAIRerank(props)
 
+/* EDIT */
+const canEdit = ref(false)
+onMounted(async () => {
+  canEdit.value = await hasRole('admin')
```
jaen22BTH: admin is just a role in our dev environment, should this not be dynamic?

kebabohvo: Lösning diskuterad med Jakke: Lista med admin roller i env variable, som sedan checkas igenom om person har en av dom i sinna roller. in progress

src/api/src/dmis_api/auth_routes.py
kebabohvo: yes removing dms-frontend-dev

src/api/src/dmis_api/auth_routes.py
```diff
@@ -104,6 +106,13 @@ async def check_auth(self, access_token: str | None = Cookie(default=None)) -> J
         if not claims:
             raise HTTPException(status_code=401)
 
+        client_roles = self._get_client_roles(claims)
+        if not client_roles:
+            raise HTTPException(
+                status_code=403,
+                detail="User does not belong to the client",
```
jaen22BTH: Don't send error messages to the frontend.

src/api/src/dmis_api/auth_routes.py
```diff
@@ -136,6 +150,23 @@ async def auth_me(self, access_token: str | None = Cookie(default=None)) -> JSON
...
+            raise HTTPException(status_code=401)
+
+        client_roles = self._get_client_roles(claims)
+
+        is_admin = any(role in client_roles for role in self.admin_roles)
+
+        return JSONResponse(
+            status_code=200,
+            content={
+                "admin": is_admin,
```
jaen22BTH, line 194: Could this not be baked in to the response in the codeexchange endpoint?

src/api/src/dmis_api/auth_routes.py
kebabohvo: fixed

src/api/src/dmis_api/auth_routes.py
kebabohvo: if we do this it breaks admin privileges in case of refresh of page, etc. We need a dedicated check for admin from what I gathered, as we check whether the user is admin to get special UI functions and meta data handling. If we only have it in codeExchange, we receive the isAdmin once, but after refresh of page or closing of page and then reopening it will not have admin access any longer.

mamz22 (approved):
Looks good to me

mamz22 (approved):

PR #520, opened 2026-05-06
Fix/confluence auth user 478

### LINKED ISSUE #478: Fix auth for Confluence connector
Confluence connector needs to tell fronted how authentication against the service should be specified.

If possible, this should be structured in an auth_user endpoint, similar to:
https://github.com/Studentprojekt-KCL/document_management_system/blob/develop/src/connectors/gitlab/src/gitlab_collector_api.py#L99 

(if not supported, input in frontend will need to be supported, and sent to this endpoint).

bjpo78:
Summary
Closes [#478](https://github.com/Studentprojekt-KCL/document_management_system/issues/478).

Confluence Cloud does not use the same OAuth redirect flow as GitLab. This PR adds a GET /auth_user endpoint on the Confluence connector that returns a JSON auth contract (manual API token + email via request headers). The connector gateway no longer assumes every downstream /auth_user response is an HTTP redirect with a Location header; it proxies redirects when appropriate and forwards JSON when the downstream response is Content-Type: application/json.

Changes
Confluence connector (collector_api)
Register GET /auth_user (same path name as OAuth connectors for consistency).

Return 200 with JSON describing manual credentials, for example:

type: api_token
method: manual
header_names: X-Confluence-Email, X-Confluence-Token
labels for frontend copy
help_url for Atlassian API token documentation
No OAuth redirect. The gateway may still send a callback-url header when proxying; this connector ignores it.

Notes
A future refactor could unify GitLab-style OAuth behind a type: oauth JSON envelope; out of scope for this PR.
dmis_api forwarding of redirects vs JSON may still warrant a follow-up if all browser flows are expected to go only through /connector/*.


src/connectors/confluence/src/confluence_service/collector_api.py
```diff
@@ -78,6 +81,24 @@ async def validation_exception_handler(self, _: Request, exc: Exception) -> JSON
             content = jsonable_encoder(errors)
         return JSONResponse(status_code=422, content=content)
 
+    async def auth_user(self) -> JSONResponse:
```
jaen22BTH, line 79: This endpoint is automatically called by the gateway, and needs to explain to the frontend how to authenticate a user in a predictable way. Also from what I understand, oauth auth to catch session token should be possible according to this:
https://developer.atlassian.com/cloud/jira/software/oauth-2-3lo-apps/

Although, I might be misunderstanding this 😅

Issue #494, opened 2026-05-05
Fix search filter options


EmEhPa:
When a user choose a filter or more, a new request will be sent including these filters so the user gets XX new search results with only these filters.

EmEhPa:
Done?

PR #527, opened 2026-05-06
Feature/frontend/filter search stuff

EmEhPa:
refers to #494

- You can filter and new requests är sent to backend and new result are sent up 
- The search matches part is scrollable and bar and filtercard is static and so the user can see it whole time
- The user can see its serach query until it enter some new query
- If security class is update d it gets refreshed in both preview drawer and search matches lsit


jaen22BTH:
Nice! Unfortunately, the filtering functionality does not seem to work on all files, for type filers.

jaen22BTH (approved):

Issue #484, opened 2026-05-04
set classification


AlexanderJarnstromBTH:
new structure for setting classification:

```json
{
   "unqiue_pointer": <pointer>,
   "security_class": <classification>
}
```

EmEhPa:
This is done

PR #531, opened 2026-05-07
Set classification #484

muaathaw:
# changes:
- Just one small change according to new structure for setting classification now we use *security_class* instead of *classification* in API body!


EmEhPa:
There is nothing to review?

PR #543, opened 2026-05-08
unit tests for dmis api auth.py #524

mamz22:


src/api/test/test_auth_routes.py
```diff
@@ -0,0 +1 @@
+"""Copyright (c) 2026, Studentprojekt Knowit Cybersecurity and Law."""
```
jaen22BTH: Should this file be here?

mamz22: No, not yet mb

jaen22BTH (approved):

PR #544, opened 2026-05-08
Feat/confluence defined fields

bjpo78:
Pushed updates to feat/confluence-defined-fields just now (within the last minute). Please refresh the PR — CI should rerun on the new commits.


jaen22BTH (commented):
I believe this solves the same issue merged in #537

PR #545, opened 2026-05-08
Feature/stochastic minor

FilipDar:
minor changes to stuff and things.


src/stochastic_analyzer/gateway/services/summarize.py
```diff
@@ -0,0 +1,93 @@
...
+    SUMMARIZE_STAGE_TWO_PROMPT,
+)
+from gateway.schemas import InputItem, SummaryResult
+
+from shared_functions.dmis_logger import dms_warning
+from shared_functions.initialisation_tools import read_env_variable
+
+
+class Summarizer:
+    """yeah buddy summarize engine"""
```
AlexanderJarnstromBTH: ye kompis!

src/stochastic_analyzer/gateway/routes.py
```diff
@@ -1,231 +1,69 @@
...
     @router.post("/md-to-pdf")
-    async def md_pdf_converter(payload: SummaryResult) -> Response:
-        """Endpoint to convert markdown to PDF."""
-        pdf: bytes = services.pdf_converter.convert(payload.summary)
+    def md_to_pdf(payload: MarkdownRequest) -> Response:
+        """Route that takes string as input and gives PDF."""
+        pdf = pdf_converter.convert(payload.markdown)
+        if pdf is None:
+            dms_warning("PDF conversion failed")
+            raise HTTPException(status_code=500)
```
AlexanderJarnstromBTH: what ya doin

src/stochastic_analyzer/gateway/services/connector.py
```diff
@@ -1,170 +1,83 @@
...
-
-            content = await self._extract(raw, display_name)
-
-            if not content:
-                dms_warning(f"Empty content after extraction for '{unique_pointer}'")
-                raise EmptyContent(display_name)
+                timeout=aiohttp.ClientTimeout(total=self.TIMEOUT),
+            ) as response:
+                response.raise_for_status()
+                data = await response.json()
```
jaen22BTH, line 46: Prob need to catch these errors to?

src/stochastic_analyzer/gateway/services/connector.py
```diff
@@ -1,170 +1,83 @@
...
+        try:
+            raw = b64decode(encoded)
+        except ValueError as err:
+            dms_warning(f"Base64 decode failed: {err}")
+            return None
 
-        Args:
-            pointers: List of unique file pointers.
+        try:
+            return Connector._converter.convert_stream(BytesIO(raw)).text_content
```
AlexanderJarnstromBTH, line 80: fat ass overhead

src/stochastic_analyzer/gateway/services/connector.py
```diff
@@ -1,170 +1,83 @@
...
+        items = []
+        for entry in data:
+            content = await asyncio.to_thread(self._extract_text, entry.get("content"))
+            if content is None:
+                continue
             items.append(
                 InputItem(
                     content=content,
-                    metadata=MetadataTemplate(unique_pointer=unique_pointer),
+                    metadata=MetadataTemplate(unique_pointer=entry.get("unique_pointer", "")),
```
jaen22BTH: It seems like a real issue if the unique_pointer does not exist, does undefined behavior not arise if we set ""?

src/stochastic_analyzer/gateway/routes.py
```diff
@@ -1,231 +1,69 @@
...
+    async def summarize(payload: PointerRequest) -> SummaryResult:
+        """Summarize one or more documents."""
+        if not payload.pointers:
+            dms_warning("summarize requires at least 1 pointer.")
+            return SummaryResult(summary="")
+
+        items = await connector.get_file_contents(payload.pointers)
+        if not items:
+            dms_warning("document retreival failure")
+            return SummaryResult(summary="")
```
jaen22BTH: As this is sent to the frontend, and we do 'missing content -> continue' in retrieve files, should there not be some indicator of why a symmery wasn't generated displayed to the user?

src/stochastic_analyzer/gateway/schemas.py
```diff
@@ -1,84 +1,35 @@
...
-
-    pointers: list[StrictStr] = Field(..., min_length=1)
+    markdown: StrictStr
 
 
 class SummaryResult(BaseModel):
-    """Output schema for a summarized document."""
+    """Schema for summary delivery."""
 
     summary: StrictStr
```
jaen22BTH: Do we really need to generate a class instance, just to define one variable, each time we return this element?

src/stochastic_analyzer/gateway/services/summarize.py
```diff
@@ -0,0 +1,93 @@
...
+                json={
+                    "model": self.model,
+                    "prompt": prompt,
+                    "system": SUMMARIZER_SYSTEM_PROMPT,
+                    "stream": False,
+                },
+                timeout=aiohttp.ClientTimeout(total=self.TIMEOUT),
+            ) as response:
+                response.raise_for_status()
+                data = await response.json()
```
jaen22BTH, line 85: Catch jsondecode error?

src/stochastic_analyzer/gateway/routes.py
```diff
@@ -1,231 +1,69 @@
...
-
     @router.post("/md-to-pdf")
-    async def md_pdf_converter(payload: SummaryResult) -> Response:
-        """Endpoint to convert markdown to PDF."""
-        pdf: bytes = services.pdf_converter.convert(payload.summary)
+    def md_to_pdf(payload: MarkdownRequest) -> Response:
+        """Route that takes string as input and gives PDF."""
+        pdf = pdf_converter.convert(payload.markdown)
+        if pdf is None:
+            dms_warning("PDF conversion failed")
```
jaen22BTH: This error is already written in the pdf_convert method, prob don't need duplicates

src/stochastic_analyzer/README.md
```diff
@@ -1,239 +1,96 @@
...
+Run the gateway locally inside a virtual python environment:
 
-**CURL Example:**
 ```bash
-curl -X POST http://localhost:8000/md-to-pdf \
-  -H "Content-Type: application/json" \
-  -d '{"summary": "# Report\n\nThis is the summary content."}' \
-  --output summary.pdf
+$ pip install -e .
+$ gateway
```
jaen22BTH: Think I missed something, but should the entrypoint not be set to something which uniquely references this, else it risks conflicts. I believe we had 'stochastic-analyzer' before.

src/stochastic_analyzer/test/test_summarize.py
```diff
@@ -0,0 +1,14 @@
+"""Unit tests for summarizer"""
```
jaen22BTH: Prob remove this file if not used

src/stochastic_analyzer/gateway/routes.py
```diff
@@ -23,47 +23,44 @@ def create_router(
...
 
-    @router.post("/summarize", response_model=SummaryResult)
-    async def summarize(payload: PointerRequest) -> SummaryResult:
+    @router.post("/summarize")
+    async def summarize(payload: PointerRequest) -> dict[str, str]:
         """Summarize one or more documents."""
         if not payload.pointers:
             dms_warning("summarize requires at least 1 pointer.")
-            return SummaryResult(summary="")
+            return {"summary": "At least one pointer is required."}
```
AlexanderJarnstromBTH: do we want to give the frontend an explanation?

src/stochastic_analyzer/gateway/routes.py
AlexanderJarnstromBTH: Wouldn't that  go against how the other services handle errors? Maybe a failure indicator should be provided, but i don't know if we should give an explanation. This might be exactly what you said, now when i read it again

src/stochastic_analyzer/gateway/routes.py
FilipDar: (╯°□°)╯︵ ┻━┻

AlexanderJarnstromBTH: maybe a 400 status code and not an explanation in the summery variable?

src/stochastic_analyzer/gateway/routes.py
jaen22BTH: Yeah no, it should definitely be a different defined behavior than that of a successful summarization. And in this case, no error should be sent. However, I feel that there has to be some defined mechanism to tell the frontend "we can not generate a summery for this type of file" and the stochastic analyzer owns the truth for this. (meaning in the code below this defined behavior must exist?)

src/stochastic_analyzer/gateway/schemas.py
```diff
@@ -1,84 +1,29 @@
...
-    """A single document item submitted for classification."""
+    """Schema for content and metadata."""
 
-    content: StrictStr = Field(..., min_length=1)
+    content: StrictStr
     metadata: MetadataTemplate
 
-    model_config = {"extra": "ignore"}
 
+class MarkdownRequest(BaseModel):
```
jaen22BTH, line 20: Is it expected that this will grow to more parameters in the future? Otherwise this feels a bit redundant. Would increase readability for humans a lot.

src/stochastic_analyzer/gateway/services/connector.py
```diff
@@ -1,170 +1,96 @@
...
+    def _extract_text(encoded: str | None) -> str | None:
+        """Decode payload and extract text. Tries markitdown first, falls back to utf-8."""
+        if encoded is None:
+            return None
+        try:
+            raw = b64decode(encoded)
+        except ValueError as err:
+            dms_warning(f"Base64 decode failed: {err}")
+            return None
+        if Connector._is_convertable(raw):
```
jaen22BTH: Has there been an active decision to make this hard assumption? If so, should there be an ADR concidering this seems like quite a significant limitation from a user perspective? (also, the docstring might need some refinement to reflect this).

src/stochastic_analyzer/gateway/schemas.py
FilipDar: In reality i didnt want to remove the old summaryresponse aswell since the FastAPI want to work with pydantic. its  the tutorial where i got this from.

https://fastapi.tiangolo.com/tutorial/body/

I kept it in schemas both for easier pydantic declaration and to make it easier to expand if necessary. 
But i can still change it if we feel like we want to skip pydantic entirely even if FastAPI seems to really want to enforce pydantic in documentatoin

src/stochastic_analyzer/gateway/services/connector.py
FilipDar: Isn't this is same approach as search engine?

AlexanderJarnstromBTH: image

AlexanderJarnstromBTH (approved):
lgtj

PR #559, opened 2026-05-11
New unit Tests for Components

muaathaw:
# This PR is intended to cover all functions and features in the components folder.

## New files

- SearchMatches.test.js
- SearchPreviewDrawer.test.js
- SearchBar.test.js
- ClassificationEditor.test.js

## New test-setup file and edit in vite.config.js to run the tests

I have commented some tests in the files cause we may not really need to test now, and was forced to disable an Eslint rule cause I couldn´t fix it in an other way.

### Update:
- Removed 12 low-value tests across 4 component test files 
- Removed duplicate of /useSearchMetadata.test.js


EmEhPa:
Great work on the test structure and coverage overall. The setup feels clean and professional. However, for this project scope, I believe we should focus more on testing behavior and core logic rather than UI details or exhaustive edge cases.

A few examples:

* Keep tests like “button disables when loading” or “emits search with correct data” since they validate real behavior/business logic.
* Consider reducing tests for static UI text, placeholders, modal titles, CSS classes, or simple rendering checks, since those are brittle and often add maintenance cost without much value.
* For views specifically: dedicated unit tests are often not necessary when views mostly compose existing components and pass props/events through. In many cases, component-level tests already cover the important behavior.

For utilities/composables, it’d be good to distinguish between:

* Pure logic worth unit testing (validation, formatting, transformations)
* Thin wrappers around APIs/rendering that may not need dedicated unit tests

The goal isn’t maximum test count, but meaningful coverage that stays maintainable and easy to reason about for a this project.

If you want, I can help propose a minimal core test matrix focused on the most important component behaviors.

PR #579, opened 2026-05-13
Fixed so that we only send access_token during regular api calls, and…

kebabohvo:
fixed pathing for cookies to be included, now we only send access_token where it is needed.
RefreshToken is sent during /auth/refresh calls
idToken is sent during /auth/logout calls for logging out properly.


mamz22 (approved):
Looks good to me

Issue #583, opened 2026-05-13
Sub Resource Integrity Attribute Missing


EmEhPa:
Risk level **Medium**

Meaning:

*The integrity attribute is missing on a script or link tag served by an external server. The integrity tag prevents an attacker who have gained access to this server from injecting a malicious content.*

Instead of using [https://cdn.jsdelivr.net/npm/marked/marked.min.js](url) (?) for converting from markdown to nice text we can use npm's marked.

EmEhPa:
Should be done #584

PR #584, opened 2026-05-13
Sub Resource Integrity Attribute Missing #583

EmEhPa:
One of the medium vulnerabilities in the owasp zap scan was "Sub Resurce Integrity Attribute Missing" for the use of <script src="https://cdn.jsdelivr.net/npm/marked/marked.min.js"></script>, the best solution use npm markedd

The visulatation is stil the same
image

Tbh, I like this code better also


mamz22 (approved):
Looks good to me

PR #585, opened 2026-05-13
Fixing cookie refresh

kebabohvo:
Fixed cookies to work with lifespan for AD tokens.
access_token & id_token Lifespan: 5 min
refresh_token lifespan: 30min

How it now works:
If a user is idle and doesn't do any api calls for let's say 12 min and then searches something, the flow is like follows:

We get an error 401 which we handle (not seen) we try to refresh tokens which happens due to 12 < 30 (refresh lifespan), the new tokens are updated from current time + 5 min for access_token & id_token, + 30 for refresh_token.

If the user is afk or idle doing nothing for 30+ minutes from previous refresh then the refresh token dies (currently) I am working on fixing this in a different branch, where we track activity, to make sure active users aren't forced logged out.


mamz22 (approved):
Looks good to me

jaen22BTH (approved):

PR #588, opened 2026-05-13
added security headers for nginx template

jesperlenn:


EmEhPa (approved):
Nice, hopefullt zap scan is smaller next time :)

Issue #602, opened 2026-05-14
Paging support in frontend


jaen22BTH:
Emma promises, 100% for sure, that paging will be complete tomorrow. :D

PR #610, opened 2026-05-14
Frontedn/paging

EmEhPa:
Soo we have paging #602

image

image

etc etc..

Vlan & Filip2 have seen it live

And like the issue says
"Emma promises, 100% for sure, that paging will be complete tomorrow. :D"

I kept my promise, its todey and not tmorrow.


AlexanderJarnstromBTH (approved):
lgtj

PR #619, opened 2026-05-14
Fix/confluence auth header

jaen22BTH:


AlexanderJarnstromBTH (approved):
fixed my marige

FilipDar (approved):
LGT jango fett

PR #629, opened 2026-05-15
removed decoding

AlexanderJarnstromBTH:


EmEhPa (approved):
I mean I believe in vlan and this looks valid

PR #644, opened 2026-05-15
Small pagining bug fixes :)

EmEhPa:
This should solve the bug for filtering while paging and go futher than 12 pages :)


jaen22BTH (approved):
Whhow, amazing!

PR #657, opened 2026-05-18
that should furfill sanitization thing in the excel

EmEhPa:
Should togetehr with other things complte this

```
Verify that output encoding for an HTTP response, HTML document, or XML document is relevant for the context required, such as encoding the relevant characters for HTML elements, HTML attributes, HTML comments, CSS, or HTTP header fields, to avoid changing the message or document structure.
```


jaen22BTH (approved):
Amazing!

Issue #662, opened 2026-05-19
Input validation


jaen22BTH:
**Input validation neeeded for Application Security Verification Standard**
 
* Input validation is needed for search input (both in frontend and search engine)
* Input validation is needed for markdown content parsed in markdown to PDF engine.
* Input validation for basic authentication is needed to prevent command injections, e.g. in SMB connector.

PR #663, opened 2026-05-19
Issue ref #662

Sjomullen:
issue ref #662


src/stochastic_analyzer/gateway/schemas.py
```diff
@@ -18,9 +18,19 @@ class InputItem(BaseModel):
...
-    markdown: StrictStr
+    """Request schema for markdown logic.
+
+    Validation rules:
+    - `markdown` must be a UTF-8 string (enforced by FastAPI's JSON decoder).
+    - Must be a string, not bool/int/float (enforced by StrictStr).
+    - Must not be empty.
+    """
+
+    markdown: StrictStr = Field(
```
jaen22BTH: Does this not risk missing the mark? Isn't real danger/sanetizing which needs to be dome more like embedded HTML, think we need to make sure this isn't parsed (atleast for markitdown there seems to be a feature like {"html": false}, not sure how it works here.

jaen22BTH (approved):

jaen22BTH (changes requested):
See comment

src/stochastic_analyzer/gateway/schemas.py
Sjomullen: didn't you say that we only needed to make sure that it was proper UTF-8 and wasn't malformed. But maybe I understood you wrong

jaen22BTH: Not sure. What we need for compliance is:
"input validation rules for checking the validity of data items against an expected structure"

Sjomullen: made some more changes now

jaen22BTH (approved):

PR #669, opened 2026-05-20
Secure cookir autribute

EmEhPa:
SO this should fulfill v.3.3.1 in the owasp excel list.

*Verify that cookies have the ‘Secure’attribute set, and if the ’__Host‑’prefix is not used for the cookie name, the ’__Secure‑’prefix must be used for the cookie name.*

The cookie is now named __Secure‑acces_token


AlexanderJarnstromBTH (approved):
lgtj

PR #673, opened 2026-05-25
Fix SharePoint connector to satisfy the connector gateway contract

FelonBall:
The SharePoint connector was missing two endpoints required by the gateway (/validate_token, /defined_fields), used the wrong HTTP method and parameter transport for /stream_files_to_index, and passed auth tokens to the Graph API with a double "Bearer " prefix.

- Add /validate_token: calls GET /me on the Graph API, returns {"valid": bool}
- Add /defined_fields: exposes the full set of fields the connector can return
- Change /stream_files_to_index from GET to POST; read subdata from the JSON body instead of a query parameter, matching the contract expected by the search engine's streaming client
- Strip the "Bearer " prefix from incoming tokens in /get_files, /stream_files_to_index, and /validate_token — the gateway forwards tokens with the prefix already included, and the interfacer prepends it again when building the Authorization header, resulting in 401s from Graph API on every authenticated call


AlexanderJarnstromBTH (approved):
lgtj

